Issue #1, opened 2020-01-23
Base HTTP interfaces


g4s8:

g4s8:
Done in #2

0crat:
Job `gh:artipie/http#1` is not assigned, can't get performer

0crat:
This job is not in scope

PR #2, opened 2020-01-23
Base HTTP interfaces

g4s8:
#1 - base HTTP interfaces:
 - request
 - response
 - front

### DIFF
--- a/src/main/java/com/artipie/http/Request.java
+++ b/src/main/java/com/artipie/http/Request.java
@@ -1,0 +1,64 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2020 Artipie
+ *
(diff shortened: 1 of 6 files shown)


src/main/java/com/artipie/http/Request.java
```diff
@@ -0,0 +1,64 @@
...
+    String line();
+
+    /**
+     * General headers, request headers and entity headers.
+     * <p>
+     * See 4.5, 5.3 and 7.1 sections of rfc2616.
+     * </p>
+     * @return Map of header values by name
+     */
+    Map<String, Iterable<String>> headers();
```
Sammers21: Let's return `Iterable<Map.Entry<String,String>>` here

Issue #7, opened 2020-02-03
TkRequest.java:83-89: Implement this class. It should...


0pdd:
The puzzle `3-21935464` from #3 has to be resolved: 

https://github.com/artipie/http/blob/661805a6d7f756d7fac34ffc5bb5544816d197dc/src/main/java/com/artipie/http/tk/TkRequest.java#L83-L89

The puzzle was created by @g4s8 on 03-Feb-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
@g4s8[/z](https://www.0crat.com/u/g4s8) please, pay attention to this issue

0crat:
There is an unrecoverable failure on my side. Please, submit it [here](https://github.com/zerocracy/farm/issues):

```
PID: 11@a89d63b9d9d0, thread: pool-778-thread-1
com.zerocracy.radars.github.ThrottledComments[74] java.lang.IllegalStateException: Can't post an identical message to artipie/http#7: @g4s8[/z](https://www.0crat.com/u/g4s8) please, pay attention to this issue
```

[0.51.0](https://github.com/zerocracy/farm): Issue: artipie/http#7, Action: `opened`

0crat:
The job [#7](https://github.com/artipie/http/issues/7) assigned to @g4s8[/z](https://www.0crat.com/u/g4s8), here is [why](https://www.0crat.com/footprint/CT2E6TK9B/3cbb73b6-b0fd-4bab-8a54-355f7b1b71b9); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be no monetary reward for this job

0crat:
[...]

PR #9, opened 2020-02-04
Request body with test

g4s8:
#7 - implemented request body publisher and unit test. Enabled qulice.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,10 @@
       <version>${junit-platform.version}</version>
       <scope>test</scope>
     </dependency>
+    <!--
+      We use RxJava for writing unit tests. It helps to convert Flow objects from JDK9 API
+      to blocking lists.
+    -->
     <dependency>
       <groupId>io.reactivex.rxjava3</groupId>
       <artifactId>rxjava</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,18 @@
       <version>${junit-platform.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.reactivex.rxjava3</groupId>
+      <artifactId>rxjava</artifactId>
+      <version>3.0.0-RC6</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.github.akarnokd</groupId>
+      <artifactId>rxjava3-jdk8-interop</artifactId>
+      <version>3.0.0-RC6</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <build>
     <pluginManagement>
@@ -84,7 +96,40 @@
           <artifactId>maven-project-info-reports-plugin</artifactId>
           <version>3.0.0</version>
         </plugin>
+        <plugin>
+          <groupId>com.qulice</groupId>
+          <artifactId>qulice-maven-plugin</artifactId>
+          <version>0.18.19</version>
+          <configuration>
+            <license>file:${project.basedir}/LICENSE</license>
+            <excludes combine.children="append">
+              <exclude>findbugs:.*</exclude>
+              <exclude>duplicatefinder:.*</exclude>
+              <exclude>dependencies:.*</exclude>
+            </excludes>
+          </configuration>
+          <executions>
+            <execution>
+              <goals>
+                <goal>check</goal>
+              </goals>
+            </execution>
+          </executions>
+        </plugin>
       </plugins>
     </pluginManagement>
   </build>
+  <profiles>
+    <profile>
+      <id>qulice</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>com.qulice</groupId>
+            <artifactId>qulice-maven-plugin</artifactId>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+  </profiles>
 </project>
--- a/src/main/java/com/artipie/http/tk/TkRequest.java
+++ b/src/main/java/com/artipie/http/tk/TkRequest.java
@@ -87,6 +87,11 @@ public final class TkRequest implements Request {
     private static final class BodySubstription implements Subscription {
 
         /**
+         * Buffer size for stream reading.
+         */
+        private static final int BUF_SIZE = 1024 * 8;
+
+        /**
          * Request input stream.
          */
         private final InputStream stream;
@@ -109,8 +114,8 @@ public final class TkRequest implements Request {
         @Override
         public void request(final long bytes) {
             try {
-                this.read((int) bytes);
-            } catch (final IOException err) {
+                this.read(bytes);
+            } catch (final IOException | IllegalArgumentException err) {
                 this.receiver.onError(err);
             }
         }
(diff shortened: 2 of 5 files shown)


amihaiemil:
@g4s8 Can you give it to me? :D

pom.xml
```diff
@@ -44,6 +44,18 @@
...
+      <artifactId>rxjava</artifactId>
+      <version>3.0.0-RC6</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.github.akarnokd</groupId>
+      <artifactId>rxjava3-jdk8-interop</artifactId>
+      <version>3.0.0-RC6</version>
+      <scope>test</scope>
+    </dependency>
```
amihaiemil: @g4s8 I get the feeling I see this kind of ``rx`` dependencies in all the artipie projects. Can you add a comment ``<!-- -->`` explaining what they are for? They are for tests, obviously, but what tools do they actually provide?

src/main/java/com/artipie/http/tk/TkRequest.java
```diff
@@ -86,6 +86,11 @@ public String line() throws IOException {
      */
     private static final class BodySubstription implements Subscription {
 
+        /**
+         * Buffer size for stream reading.
+         */
+        private static final int BUF_SIZE = 1024 * 8;
```
amihaiemil: @g4s8 I guess this is the ``max`` buffersize. Should we name it as such?

src/main/java/com/artipie/http/tk/TkRequest.java
```diff
@@ -108,16 +113,52 @@ public String line() throws IOException {
...
         public void cancel() {
-            throw new UnsupportedOperationException(
-                String.format("cancel not implemented: %s/%s", this.stream, this.receiver)
-            );
+            try {
+                this.stream.close();
+            } catch (final IOException err) {
+                this.receiver.onError(err);
+            }
+        }
```
amihaiemil: @g4s8 Can we have a test for this method as well? For instance, what happens if ``cancel()`` is called twice, or if ``request(bytes)`` is called after ``cancel()``?

src/main/java/com/artipie/http/tk/TkRequest.java
```diff
@@ -108,16 +113,52 @@ public String line() throws IOException {
...
+
+        /**
+         * Read bytes from stream into receiver.
+         * @param bytes Amount of bytes to read
+         * @throws IOException On stream error
+         */
+        private void read(final long bytes) throws IOException {
+            if (bytes <= 0) {
+                throw new IllegalArgumentException(String.format("can't request %d bytes", bytes));
+            }
```
amihaiemil: @g4s8 I would put this ``if`` in the ``request()`` method, right before calling this method.

src/test/java/com/artipie/http/tk/TkRequestTest.java
```diff
@@ -77,4 +80,30 @@ void readsEmptyHeaders() throws Exception {
             Matchers.anEmptyMap()
         );
     }
+
+    @Test
+    void readsBody() throws Exception {
+        final byte[] body = new byte[]{0x00, 0x01, 0x02, 0x03};
```
amihaiemil: @g4s8 Is it possible to write this test by using a String, or some chars? Something more meaningful than actual hardcoded bytes :D

src/main/java/com/artipie/http/tk/TkRequest.java
```diff
@@ -108,16 +113,52 @@ public String line() throws IOException {
...
+                }
+                final int read = this.stream.read(buf, 0, len);
+                total += read;
+                if (read == -1) {
+                    this.stream.close();
+                    this.receiver.onComplete();
+                    break;
+                } else {
+                    for (int pos = 0; pos < read; ++pos) {
+                        this.receiver.onNext(buf[pos]);
```
amihaiemil: @g4s8 ``onNext`` what? What does this method do actually? :D

src/main/java/com/artipie/http/tk/TkRequest.java
```diff
@@ -108,16 +113,52 @@ public String line() throws IOException {
...
+         */
+        private void read(final long bytes) throws IOException {
+            if (bytes <= 0) {
+                throw new IllegalArgumentException(String.format("can't request %d bytes", bytes));
+            }
+            final byte[] buf = new byte[TkRequest.BodySubstription.BUF_SIZE];
+            long total = 0;
+            while (total < bytes) {
+                final int len;
+                if (total + TkRequest.BodySubstription.BUF_SIZE <= bytes) {
```
amihaiemil: @g4s8 I think this should be ``>=`` instead of ``<=``. So if the maximum buf size is exceeded, then ``len`` takes ``buf_size``. But I am not sure :D 

Ugly algorithm... Isn't it possible to write this algorithm with class ``IOUtils`` from Apache Commons or something?

0crat:
This pull request [#9](https://github.com/artipie/http/issues/9) is assigned to @amihaiemil[/z](https://www.0crat.com/u/amihaiemil), here is [why](https://www.0crat.com/footprint/CT2E6TK9B/4a6e1901-67ce-44c5-babd-5c99da154d70); the budget is 15 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§27](http://www.zerocracy.com/policy.html#27) and when you decide to accept the changes, inform @g4s8[/z](https://www.0crat.com/u/g4s8) (the architect) right in this ticket; if you decide that this PR should not be accepted ever, also inform the architect; this [blog post](http://www.yegor256.com/2015/02/09/serious-code-reviewer.html) will help you understand what is expected from a code reviewer; there will be a monetary reward for this job

0crat:

src/main/java/com/artipie/http/tk/TkRequest.java
g4s8: @amihaiemil I don't think so - this is just a buffer size. The buffer is allocated with this size and used to read bytes from stream into this buffer. The buffer's size is always 8K.

src/main/java/com/artipie/http/tk/TkRequest.java
g4s8: @amihaiemil `onNext` is a part of JDK9 [Flow API](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html) you can check the javadoc here: https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.Subscriber.html#onNext-T-

src/main/java/com/artipie/http/tk/TkRequest.java
g4s8: @amihaiemil yes, it looks ugly, maybe I'll extract it into separate library. I didn't find implementations to convert Flow to Streams. The logic is correct here: if requested amount of bytes to read is greater than already read bytes (total) + buffer size, then we read full buffer, else we need to read only missing chunk of bytes (requested amount - already read).

src/test/java/com/artipie/http/tk/TkRequestTest.java
g4s8: @amihaiemil I don't think it's correct to use string for bytes testing

amihaiemil: @g4s8 I was thinking of ``String.getBytes()`` rather than declaring a raw byte array :)

src/main/java/com/artipie/http/tk/TkRequest.java
amihaiemil: @g4s8 Can you add this algorithm's description (strep by step) in a JavaDoc over the method? So people can figure it out easier :D

g4s8:
@amihaiemil done, please check new commit and my answers https://github.com/artipie/http/pull/9/commits/de96b7ec3e7edd1daf99b6323ec2dd258dc8d82b

amihaiemil:
@g4s8 the new commit looks good. Just see my latest 2 comments: [here](https://github.com/artipie/http/pull/9#discussion_r374616207) and [here](https://github.com/artipie/http/pull/9#discussion_r374616599).

src/main/java/com/artipie/http/tk/TkRequest.java
g4s8: @amihaiemil added javadoc

src/test/java/com/artipie/http/tk/TkRequestTest.java
g4s8: @amihaiemil I still don't understand why we should use string here. I think it may confuse the reader, because if we use array of bytes here, it becomes obvious that this class uses bytes for data transferring. Otherwise, the reader may misunderstand that class assumes some string to be encoded into byte array.

0crat:
Job was finished in 3 hours, bonus for fast delivery is possible (see [§36](http://www.zerocracy.com/policy.html#36))

0crat:
Order was finished: +20 point(s) just awarded to @amihaiemil[/z](https://www.0crat.com/u/amihaiemil)

0crat:

PR #18, opened 2020-02-05
Api improvements

### LINKED ISSUE #19: Request / response redesign
Redesign `Request`, `Response` and `Slice` interfaces to make it reactive.

Sammers21:
* `line`, `headers`, `body` should not throw any exceptions
* response body should be a publisher, not a subscriber

fixes #19


g4s8:
@0crat assign me

g4s8:
@Sammers21 @yegor256 I don't like these names: `Destination`, `renderIt`, `renderTo`, it doesn't describe objects behavior. I'm suggesting to name use:
 - `Connection` with `accept(status, headers, body)` method; we're asking connection to accept our response data
 - `Response` with `send(connection)` method; we're asking response to send itself to connection

The responsibility of connection: accepting response data. It will be implemented by HTTP server.
The responsibility of response is to send itself via connection: generate response data and ask connection to accept it.

The code may look:
```java
class MavenRepo implements Slice {
  Response accept(line, headers, body) {
    return new StatusResponse(404);
  }
}

class StatusResponse implements Response {
  
  StatusResponse(status) {
    this.status = status;
  }

  void send(connection) {
    connection.accept(this.status, emty, empty);
  }
}

class HttpConnection implements Connection {
  void accept(code, headers, body) {
    transport.send(code, headers, body);
  }
}
```

yegor256:
@g4s8 @Sammers21 I like the suggested design

src/main/java/com/artipie/http/Destination.java
```diff
@@ -21,48 +21,26 @@
...
-     * @return Map of header values by name
-     * @throws IOException On IO failure
+     * Render the destination.
+     * @param code The http status code.
+     * @param headers The http request headers.
+     * @param body The http response body.
      */
-    Map<String, Iterable<String>> headers() throws IOException;
+    void renderIt(int code,
+        List<Map.Entry<String, String>> headers,
```
g4s8: @Sammers21 I'd keep it as `Map<String, Iterable<String>>` for two reasons:
 1 - list provides API to modify itself, it's not acceptable in case of response. Response headers should not be changed after construction
 2 - readability - it would be easier to understand that headers is a mapping with one-to-many relation when using `Map`

src/main/java/com/artipie/http/Slice.java
```diff
@@ -39,9 +40,13 @@
...
-     * @return Future response
+     * Respond to a http request.
+     * @param line The request line
+     * @param headers The request headers
+     * @param body The request body
+     * @return The response.
      */
-    CompletableFuture<Response> respond(Request req);
+    Response response(String line,
+        List<Map.Entry<String, String>> headers,
```
g4s8: @Sammers21 same here about headers. Also, it would be hard to find header by name using list of entries

src/main/java/com/artipie/http/Response.java
```diff
@@ -35,30 +33,9 @@
...
-     */
-    Map<String, Iterable<String>> headers();
-
-    /**
-     * Response body flow.
-     * <p>
-     * See 7.2 section of rfc2616.
-     * </p>
-     * @return Bytes flow
+     * Render response.
```
g4s8: @Sammers21 why did you remove javadoc links to HTTP spec from `Response` class?

0crat:
> @0crat assign me ([here](https://github.com/artipie/http/issues/18#issuecomment-582421073))

@g4s8 This pull request [#18](https://github.com/artipie/http/issues/18) is assigned to @g4s8[/z](https://www.0crat.com/u/g4s8), here is [why](https://www.0crat.com/footprint/CT2E6TK9B/bdda064c-c937-4b3d-9dd7-7d4c390abcf0); the budget is 15 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§27](http://www.zerocracy.com/policy.html#27) and when you decide to accept the changes, inform @g4s8[/z](https://www.0crat.com/u/g4s8) (the architect) right in this ticket; if you decide that this PR should not be accepted ever, also inform the architect; this [blog post](http://www.yegor256.com/2015/02/09/serious-code-reviewer.html) will help you understand what is expected from a code reviewer; there will be no monetary reward for this job

0crat:

src/main/java/com/artipie/http/Destination.java
Sammers21: > 1- list provides API to modify itself, it's not acceptable in case of response. Response headers should not be changed after construction

Why do you think that `Map` is acceptable then?

> 2 - readability - it would be easier to understand that headers is a mapping with one-to-many relation when using Map

It is subjective. I would say `List<Map.Entry<String, String>>`  better coincide with the HTTP specification, hence more understandable and clear.

g4s8: @Sammers21 it makes sense, let's use `Iterable<Map.Entry<String, String>>` then

Sammers21: @g4s8, `Iterable` also allow modifications

g4s8: @Sammers21 you can't: https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html 
All you can do with `Iterable` is iterating over its items

Sammers21: Iterators allow the caller to remove elements from the underlying collection during the iteration with well-defined semantics

g4s8: @Sammers21 but the default implementation of `remove()` throws `UnsupportedOperationException`: https://docs.oracle.com/javase/8/docs/api/java/util/Iterator.html#remove--

Sammers21: @g4s8, what is wrong with using `List` then?  We can use a `List` implementation which will throw an exception on modification attempt

g4s8: @Sammers21 `List` is too specific for this case. We don't need the majority of `List` methods in request headers parsing, e.g. we don't need `add()`, `remove()`, `get(8)`, `clear()` and others. All we need here is iterating over headers. Maybe it would be better to create custom `Headers` type with methods like `headers.values("header-name")`

src/main/java/com/artipie/http/Connection.java
```diff
@@ -21,48 +21,25 @@
...
-     * </p>
-     * @return Bytes flow
-     * @throws IOException On IO failure
+     * Render the destination.
+     * @param code The http status code.
+     * @param headers The http request headers.
+     * @param body The http response body.
      */
-    Flow.Publisher<Byte> body() throws IOException;
+    CompletableFuture<Void> respond(int code,
```
g4s8: @Sammers21 I think it should be names `accept()`, see https://github.com/artipie/http/pull/18#issuecomment-582428950

Also, I don't think we need `CompletableFuture` here, since when we're asking  connection to accept response data, we don't care how and when exactly connection will accept it

src/main/java/com/artipie/http/Response.java
```diff
@@ -35,30 +33,10 @@
...
-     * See 7.2 section of rfc2616.
-     * </p>
-     * @return Bytes flow
+     * Send the response.
+     *
+     * @param connection Connection to send the response to
+     * @return Completion or error signal.
      */
-    Flow.Subscriber<Byte> body();
+     CompletableFuture<Void> send(Connection connection);
```
g4s8: @Sammers21 same here: when we send response via connection, we don't care when the connection will accept it. It's not async operation, because response should not perform blocking operations on `send()`, all body flow will be published via `Publisher<>` body interface.

src/main/java/com/artipie/http/Connection.java
Sammers21: `accept` is too general

0crat:

0crat:
Order was finished: +15 point(s) just awarded to @g4s8[/z](https://www.0crat.com/u/g4s8)

PR #13, opened 2020-02-04
Request line helper object

g4s8:
#12 - added request line helper object with unit tests

PR #23, opened 2020-02-06
add Request Line

Sammers21:
Request line would simplify request line formation by a server

### DIFF
--- a/src/main/java/com/artipie/http/RequestLine.java
+++ b/src/main/java/com/artipie/http/RequestLine.java
@@ -1,0 +1,65 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2020 Artipie
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


src/main/java/com/artipie/http/RequestLine.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public RequestLine(final String method, final String uri, final String version) {
+        this.method = method;
+        this.uri = uri;
+        this.version = version;
+    }
+
+    @Override
+    public String toString() {
+        return String.format("%s %s %s\r\n", this.method, this.uri, this.version);
```
g4s8: @Sammers21 we don't need these line-endings here `\r\n`, it's the separator for request line and headers

Sammers21: according to spec: `Request-Line   = Method SP Request-URI SP HTTP-Version CRLF`

g4s8: @Sammers21 updated in https://github.com/artipie/http/pull/13/commits/6e1d00a2638378a143df47aeae7854acf65688ec

yegor256:
@Sammers21 no unit test?

Sammers21:
@yegor256, just created a test.

0crat:
This job is not in scope

0crat:
Job `gh:artipie/http#23` is not assigned, can't get performer

PR #49, opened 2020-02-20
Implemented headers helper

g4s8:
#48 - added two header helper objects: `RqHeaders` as `List<String>` and `RqHeaders.Single` as `Text`.

### DIFF
--- a/src/main/java/com/artipie/http/rq/RqHeaders.java
+++ b/src/main/java/com/artipie/http/rq/RqHeaders.java
@@ -1,0 +1,117 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2020 Artipie
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package com.artipie.http.rq;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+import org.cactoos.Text;
+import org.cactoos.collection.Filtered;
+import org.cactoos.list.ListEnvelope;
+import org.cactoos.list.Mapped;
+
+/**
+ * Request headers.
+ * <p>
+ * Request header values by name from headers.
+ * Usage (assume {@link com.artipie.http.Slice} implementation):
+ * <pre><code>
+ *  Response response(String line, Iterable&lt;Map.Entry&lt;String, String&gt;&gt; headers,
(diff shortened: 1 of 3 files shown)


paulodamaso:
@g4s8 please assign me to this review

0crat:
@g4s8[/z](https://www.0crat.com/u/g4s8) not enough funds available in the project, can't set budget of job [#49](https://github.com/artipie/http/issues/49), see [§21](http://www.zerocracy.com/policy.html#21); @paulodamaso[/z](https://www.0crat.com/u/paulodamaso) will get no money on completion; in order to fix that, add funds to the project and assign the job again

src/test/java/com/artipie/http/rq/RqHeadersTest.java
```diff
@@ -0,0 +1,102 @@
...
+ * @since 0.4
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
+ */
+public final class RqHeadersTest {
+
+    @Test
+    void findsAllHeaderValues() {
+        final String first = "1";
+        final String second = "2";
+        MatcherAssert.assertThat(
```
paulodamaso: @g4s8 Please add a failure message to this assertion

src/test/java/com/artipie/http/rq/RqHeadersTest.java
```diff
@@ -0,0 +1,102 @@
...
+                    new IsEqual<>(second)
+                )
+            )
+        );
+    }
+
+    @Test
+    void findSingleValue() {
+        final String value = "text/plain";
+        MatcherAssert.assertThat(
```
paulodamaso: @g4s8 Please add a failure message to this assertion

src/main/java/com/artipie/http/rq/RqHeaders.java
```diff
@@ -0,0 +1,117 @@
...
+ * Header names are case-insensitive, according to
+ * <a href="https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2">RFC2616 SPEC</a>:
+ * <br>
+ * &gt; Each header field consists of a name followed by a colon (":") and the field value.
+ * <br>
+ * &gt; Field names are case-insensitive
+ * </p>
+ * @since 0.4
+ */
+public final class RqHeaders extends ListEnvelope<String> implements List<String> {
```
paulodamaso: @g4s8 I just see one inconsistency here: if `RqHeaders` contract follows `ListEnvelope<String>`, we expect that a `Single` header to be a `String`. In this case, you've made it `Text`. To avoid confusion and keep similar constructs being represented by the same classes, make `RqHeaders` contract follows `ListEnvelope<`Text>` and make `Single` implement `Text`

0crat:

src/test/java/com/artipie/http/rq/RqHeadersTest.java
g4s8: @paulodamaso why? This method has descriptive name (I think) and has the only assertion, why do we need to duplicate method name in assertion messag?

src/test/java/com/artipie/http/rq/RqHeadersTest.java
g4s8: @paulodamaso same here

src/main/java/com/artipie/http/rq/RqHeaders.java
g4s8: @paulodamaso I don't agree: `List<String>` is a source object of multiple strings, but `Text` is a source object for single string, so the semantic of both classes `List<String> -> String` and `Text -> String` are similar, don't you think?

src/test/java/com/artipie/http/rq/RqHeadersTest.java
paulodamaso: @g4s8 We are all using failure messages in the assertions (at least I ask the reviewers to put them), so it's a good idea to make the code uniform. Also, I think that a human friendly message won't hurt in this case

src/main/java/com/artipie/http/rq/RqHeaders.java
paulodamaso: @g4s8 I don't see this way, it is strange to me that a single header is represented by a `Text` object and a collection of Headers is represented by a `List<String>` one. In my opinion we should chose one way to represent a single header (`Text` or `String`, doesn't really matter) and use a collection of this this very same type to represent a collection of headers. In fact, maybe we should have `Header` object and have `Headers` collective  as being a collection of `Header`, but I think that this may be outside of the scope of this issue

g4s8:
@paulodamaso thanks, fixed two comments, but don't agree with `Text` usage

0crat:
Job [#49](https://github.com/artipie/http/issues/49) is not in the agenda of @paulodamaso[/z](https://www.0crat.com/u/paulodamaso), can't retrieve data and time of add

0crat:
Order was finished: +30 point(s) just awarded to @paulodamaso[/z](https://www.0crat.com/u/paulodamaso)

0crat:

PR #130, opened 2020-04-21
Extend AsyncResponse with constructor which accepts a single

Sammers21:
Improves AsyncReponse user experience.

### DIFF
--- a/src/main/java/com/artipie/http/async/AsyncResponse.java
+++ b/src/main/java/com/artipie/http/async/AsyncResponse.java
@@ -25,6 +25,8 @@ package com.artipie.http.async;
 
 import com.artipie.http.Connection;
 import com.artipie.http.Response;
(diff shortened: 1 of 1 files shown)


g4s8:
@Sammers21 why do we need it? I think it's better to use standard JDK classes for library API, e.g. `CompletableFuture` or `CompletionStage`

Sammers21:
@g4s8 , Yes, we stick with `CompletableFuture` and `CompletionStage`, but I want to avoid an intermediate conversion operation. The avoidance improves resulted in user experience. Especially when this class is used in JRuby.

0crat:
Job `gh:artipie/http#130` is not assigned, can't get performer

Sammers21:
@g4s8 , can you release a new version?

g4s8:
@rultor release, tag=`0.8.1`

Issue #203, opened 2020-07-20
GroupSlice: Wrong order of result


g4s8:
GroupSlice may fail with error:
```
artipie         | [INFO] 2020-07-20 13:04:16,093 HttpClient@19dbfba7-26 com.artipie.http.DockerRoutingSlice - Failure: java.util.concurrent.CompletionException: java.lang.IllegalStateException: Wrong order of result
artipie         | 	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(Unknown Source)
artipie         | 	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(Unknown Source)
artipie         | 	at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(Unknown Source)
artipie         | 	at java.base/java.util.concurrent.CompletableFuture.postComplete(Unknown Source)
artipie         | 	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(Unknown Source)
artipie         | 	at io.reactivex.internal.observers.ConsumerSingleObserver.onError(ConsumerSingleObserver.java:46)
artipie         | 	at io.reactivex.internal.operators.maybe.MaybeSwitchIfEmptySingle$SwitchIfEmptyMaybeObserver$OtherSingleObserver.onError(MaybeSwitchIfEmptySingle.java:122)
artipie         | 	at io.reactivex.internal.operators.single.SingleDelayWithCompletable$OtherObserver.onError(SingleDelayWithCompletable.java:64)
artipie         | 	at io.reactivex.internal.operators.single.SingleFlatMapCompletable$FlatMapCompletableObserver.onError(SingleFlatMapCompletable.java:97)
artipie         | 	at io.reactivex.subjects.SingleSubject.onError(SingleSubject.java:153)
artipie         | 	at hu.akarnokd.rxjava2.interop.SingleInterop.lambda$fromFuture$2(SingleInterop.java:79)
artipie         | 	at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(Unknown Source)
artipie         | 	at java.base/java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(Unknown Source)
artipie         | 	at java.base/java.util.concurrent.CompletableFuture.postComplete(Unknown Source)
artipie         | 	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(Unknown Source)
[...]

PR #204, opened 2020-07-21
Unit test for group slice

g4s8:
#203 - added a unit test for group slice to verify that slice returns expected status code from grouped items.

### DIFF
--- a/src/main/java/com/artipie/http/async/AsyncResponse.java
+++ b/src/main/java/com/artipie/http/async/AsyncResponse.java
@@ -28,6 +28,7 @@ import com.artipie.http.Response;
 import hu.akarnokd.rxjava2.interop.SingleInterop;
 import io.reactivex.Single;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ExecutionException;
 
 /**
  * Async response from {@link CompletionStage}.
@@ -59,5 +60,23 @@ public final class AsyncResponse implements Response {
     @Override
     public CompletionStage<Void> send(final Connection connection) {
         return this.future.thenCompose(rsp -> rsp.send(connection));
+    }
+
+    // @checkstyle ReturnCountCheck (15 lines)
+    @Override
+    @SuppressWarnings("PMD.OnlyOneReturn")
+    public String toString() {
+        try {
+            return String.format(
+                "%s: %s",
+                this.getClass().getSimpleName(),
+                this.future.toCompletableFuture().get()
+            );
+        } catch (final InterruptedException ignore) {
+            Thread.currentThread().interrupt();
+            return "";
+        } catch (final ExecutionException err) {
+            throw new IllegalStateException(err);
+        }
     }
 }
--- a/src/main/java/com/artipie/http/async/AsyncResponse.java
+++ b/src/main/java/com/artipie/http/async/AsyncResponse.java
@@ -28,7 +28,6 @@ import com.artipie.http.Response;
 import hu.akarnokd.rxjava2.interop.SingleInterop;
 import io.reactivex.Single;
 import java.util.concurrent.CompletionStage;
-import java.util.concurrent.ExecutionException;
 
 /**
  * Async response from {@link CompletionStage}.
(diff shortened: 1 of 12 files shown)


src/main/java/com/artipie/http/group/GroupConnection.java
```diff
@@ -70,7 +71,7 @@
         synchronized (this.results) {
             return this.results.complete(
                 this.pos,
-                new GroupResult(status, headers, body),
+                new GroupResult(status, headers, new OneTimePublisher<>(body)),
```
olegmoz, line 74: @g4s8 do not really understand why you want to use `OneTimePublisher` outside of tests

src/test/java/com/artipie/http/group/GroupSliceTest.java
```diff
@@ -0,0 +1,91 @@
...
+    @Test
+    void groupFromMultipleSources() {
+        // @checkstyle MagicNumberCheck (10 lines)
+        MatcherAssert.assertThat(
+            new GroupSlice(
+                new SliceWithDelay(new SliceSimple(StandardRs.NOT_FOUND), Duration.ofMillis(250)),
+                new SliceWithDelay(new SliceSimple(StandardRs.NOT_FOUND), Duration.ofMillis(50)),
+                new SliceWithDelay(new SliceSimple(StandardRs.OK), Duration.ofMillis(150)),
+                new SliceWithDelay(new SliceSimple(StandardRs.NOT_FOUND), Duration.ofMillis(200))
+            ),
```
olegmoz, line 56: @g4s8 I'd suggest to add failing slice and slice that never responds to the mix, to see if `GroupSlice` replies as soon as there is an `OK` response

g4s8:
@olegmoz thanks, I've fixed all your comments. Also, `GroupResults` class was fixed after improving unit test.

src/main/java/com/artipie/http/async/AsyncResponse.java
```diff
@@ -60,4 +61,22 @@ public AsyncResponse(final CompletionStage<? extends Response> future) {
...
+                "%s: %s",
+                this.getClass().getSimpleName(),
+                this.future.toCompletableFuture().get()
+            );
+        } catch (final InterruptedException ignore) {
+            Thread.currentThread().interrupt();
+            return "";
+        } catch (final ExecutionException err) {
+            throw new IllegalStateException(err);
+        }
```
olegmoz, line 80: @g4s8 this code looks dangerous to me. Trying to log such object in production code may require it to block or crash. I think that common contract for `toString()` is to be non-blocking and be safe as possible

g4s8: @olegmoz yes, I thought about it, but it's hard to choose how we should use this `toString` method. There are two points about it:
 - it's dangerous for production code and may lead to unpredictable behavior
 - unit test doesn't show what's wrong with response without this method (it's just showing something like "Expected response with status OK, but was AsyncResponse@0x1234"

Maybe we can find a compromise by checking if we're running in JUnit environment and use dangerous version of `toString` in that case, WDYT?

olegmoz: @g4s8 checking inside `toString` will make it aware of JUnit. Can't this reporting problem in tests be solved with custom output in `Matchers`? Not sure about it.

Another solution for this issue would be to check if future `isDone` inside the method and print the value. Print `uncomplete` or `failed: <reason>` otherwise. The future should always be complete when matching has happened.

g4s8: @olegmoz I replaced it with `this.future.toString()`

g4s8:
@olegmoz thanks, please check new commit https://github.com/artipie/http/pull/204/commits/d6a7fcbadfbd82370d8c87aa9db3a584fe97c568

0crat:
Job was finished in 25 hours, bonus for fast delivery is possible (see [§36](http://www.zerocracy.com/policy.html#36))

0crat:
Job audit: performer @paulodamaso[/z](https://www.0crat.com/u/paulodamaso)[/z](https://www.0crat.com/u/paulodamaso) didn't make CR comments

Issue #206, opened 2020-07-30
HTTP groupped slice should broadcast only read requests


g4s8:
`GroupSlice` should send `PUT`, `POST`, `PATCH` and `DELETE` requests only to first `Slice` target, not for all `Slices`. Only `GET`, `HEAD`, `OPTIONS`, `TRACE` and `CONNECT` should be sent as broadcast requests.

PR #207, opened 2020-07-30
Fixed group slice

g4s8:
#206 - send only read requests to group slices, all write/modify requests are going to first slice only

### DIFF
--- a/src/main/java/com/artipie/http/group/GroupSlice.java
+++ b/src/main/java/com/artipie/http/group/GroupSlice.java
@@ -25,11 +25,15 @@ package com.artipie.http.group;
 
 import com.artipie.http.Response;
 import com.artipie.http.Slice;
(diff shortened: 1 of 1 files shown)


src/main/java/com/artipie/http/group/GroupSlice.java
```diff
@@ -40,6 +44,17 @@
  */
 public final class GroupSlice implements Slice {
```
olenagerasimova, line 45: @g4s8 I'd suggest adding todo to write test for `GroupSlice`

g4s8: @olenagerasimova thanks, but the test for `GroupSlice` was introduced in another PR already

0crat:
Job `gh:artipie/http#207` is not assigned, can't get performer

PR #215, opened 2020-08-06
#214 - Better types and more constructors header matchers

### LINKED ISSUE #214: Better headers types in `RsHasHeaders` and `ResponseMatcher`
`RsHasHeaders` accepts only `Collection` of headers in constructor and does not accept `Iterable` - it does not make it compatible with `Headers` interfaces which is widely used.
`ResponseMatcher` does not has constructors supporting headers as `Iterable`, only arrays are accepted. This again makes it unusable in some cases

olegmoz:
Closes #214 
Better types and more constructors in `RsHasHeaders` and `ResponseMatcher`.

### DIFF
--- a/src/main/java/com/artipie/http/hm/ResponseMatcher.java
+++ b/src/main/java/com/artipie/http/hm/ResponseMatcher.java
@@ -24,9 +24,9 @@
 package com.artipie.http.hm;
 
 import com.artipie.http.Response;
-import com.artipie.http.headers.Header;
 import com.artipie.http.rs.RsStatus;
 import java.nio.charset.Charset;
+import java.util.Map;
 import org.cactoos.list.ListOf;
 import org.hamcrest.Matcher;
 import org.hamcrest.core.AllOf;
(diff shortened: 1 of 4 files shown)


olegmoz:
@g4s8 could you please review this PR?

src/main/java/com/artipie/http/hm/ResponseMatcher.java
```diff
@@ -39,11 +39,37 @@
...
+    }
+
+    /**
+     * Ctor.
+     * @param status Expected status
+     * @param body Expected body
+     * @param headers Expected headers
+     */
+    @SafeVarargs
+    public ResponseMatcher(
```
g4s8, line 68: @olegmoz I'd use single order for constructor parameters: now we have `status, body, headers` and `status, headers, body`. I think second option would be more correct here.

olegmoz: @g4s8 thanks, fixed parameters order for this method

g4s8:
@olegmoz thanks

0crat:
Job `gh:artipie/http#215` is not assigned, can't get performer

Issue #196, opened 2020-07-08
Status code classification in `RsStatus`


olegmoz:
[RFC](https://tools.ietf.org/html/rfc7231) breaks down codes into categories by first digit. It would be useful to have these categories in `RsStatus`. For example we may have methods like `bool clientError()` for 4xx, `bool serverError()` for 5xx and `bool error() { return clientError() || serverError(); }`.

Originated in https://github.com/artipie/artipie/pull/283#discussion_r451498806

0crat:
@g4s8[/z](https://www.0crat.com/u/g4s8), @olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) please, pay attention to this issue

g4s8:
@olegmoz sounds reasonable

g4s8:
@rultor release, tag=`0.15.6`

0crat:
Job `gh:artipie/http#196` is not assigned, can't get performer

PR #217, opened 2020-08-12
#196 classification methods added

genryxy:
For #196 classification methods `bool isClientError()`, `bool isServerError()`, `bool isError()` added

### DIFF
--- a/src/main/java/com/artipie/http/rs/RsStatus.java
+++ b/src/main/java/com/artipie/http/rs/RsStatus.java
@@ -153,6 +153,33 @@ public enum RsStatus {
     }
 
     /**
+     * Checks whether the RsStatus is a client error.
+     * @return True if the RsStatus is a client error, otherwise - false.
+     * @since 0.16
+     */
+    public boolean isClientError() {
+        return this.string.charAt(0) == '4';
+    }
+
+    /**
+     * Checks whether the RsStatus is a server error.
+     * @return True if the RsStatus is a server error, otherwise - false.
+     * @since 0.16
+     */
+    public boolean isServerError() {
+        return this.string.charAt(0) == '5';
+    }
+
+    /**
+     * Checks whether the RsStatus is an error.
+     * @return True if the RsStatus is an error, otherwise - false.
+     * @since 0.16
+     */
+    public boolean isError() {
+        return this.isClientError() || this.isServerError();
+    }
+
+    /**
      * Searches {@link RsStatus} instance by response code.
      * @since 0.11
      */
--- a/src/main/java/com/artipie/http/rs/RsStatus.java
+++ b/src/main/java/com/artipie/http/rs/RsStatus.java
@@ -153,21 +153,48 @@ public enum RsStatus {
     }
 
     /**
-     * Checks whether the RsStatus is a client error.
-     * @return True if the RsStatus is a client error, otherwise - false.
+     * Checks whether the RsStatus is an informational group (1xx).
+     * @return True if the RsStatus is 1xx, otherwise - false.
      * @since 0.16
      */
-    public boolean isClientError() {
-        return this.string.charAt(0) == '4';
+    public boolean information() {
+        return this.firstSymbol('1');
+    }
+
+    /**
+     * Checks whether the RsStatus is a successful group (2xx).
+     * @return True if the RsStatus is 2xx, otherwise - false.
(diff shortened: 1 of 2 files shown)


genryxy:
@olegmoz check please

src/main/java/com/artipie/http/rs/RsStatus.java
```diff
@@ -152,6 +152,33 @@ public String code() {
...
+    }
+
+    /**
+     * Checks whether the RsStatus is an error.
+     * @return True if the RsStatus is an error, otherwise - false.
+     * @since 0.16
+     */
+    public boolean isError() {
+        return this.isClientError() || this.isServerError();
+    }
```
olegmoz, line 180: @genryxy in this project we do not use popular getter/setter naming convention. We are naming methods as verbs if the do some action and as nouns if they return some value. So here it would be more along the lines to name methods `error`, `clientError`, `serverError`.

src/main/java/com/artipie/http/rs/RsStatus.java
```diff
@@ -152,6 +152,33 @@ public String code() {
...
+
+    /**
+     * Checks whether the RsStatus is a server error.
+     * @return True if the RsStatus is a server error, otherwise - false.
+     * @since 0.16
+     */
+    public boolean isServerError() {
+        return this.string.charAt(0) == '5';
+    }
```
olegmoz, line 171: @genryxy I'd suggest to extract common code for checking first symbol into private method

src/main/java/com/artipie/http/rs/RsStatus.java
```diff
@@ -152,6 +152,33 @@ public String code() {
         return this.string;
     }
```
olegmoz, line 154: @genryxy please add all classes of status codes from [RFC](https://tools.ietf.org/html/rfc7231), it's 6.x points in the document

src/test/java/com/artipie/http/rs/RsStatusTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    void createsNotError() {
+        final RsStatus success = RsStatus.OK;
+        final RsStatus rscontinue = RsStatus.CONTINUE;
+        final RsStatus found = RsStatus.FOUND;
+        Assertions.assertFalse(success.isError());
+        Assertions.assertFalse(rscontinue.isError());
+        Assertions.assertFalse(found.isError());
+    }
```
olegmoz, line 59: @genryxy we are using hamcrest matchers for all tests in the project. Also if test has more then assertion, that it needs an assertion message

src/test/java/com/artipie/http/rs/RsStatusTest.java
```diff
@@ -0,0 +1,61 @@
...
+        Assertions.assertFalse(status.isClientError());
+        Assertions.assertTrue(status.isError());
+        Assertions.assertTrue(status.isServerError());
+    }
+
+    @Test
+    void createsNotError() {
+        final RsStatus success = RsStatus.OK;
+        final RsStatus rscontinue = RsStatus.CONTINUE;
+        final RsStatus found = RsStatus.FOUND;
```
olegmoz, line 55: @genryxy it would be better to use `@ParameterizedTest` here, to run same test for several status codes

src/test/java/com/artipie/http/rs/RsStatusTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test for {@link RsStatus}.
+ *
+ * @since 0.16
+ */
+final class RsStatusTest {
+    @Test
+    void createsClientError() {
```
olegmoz, line 36: @genryxy test method names do not make much sense as `RsStatus` does not create anything. I'd rather name them `shouldBe[ClientError/ServerError/Error]` or `test[ClientError/ServerError/Error]`. WDYT?

genryxy: @olegmoz thanks, I think you're right

genryxy:
@olegmoz thanks, all fixed

olegmoz:
@g4s8 good to merge

0crat:
Job `gh:artipie/http#217` is not assigned, can't get performer

PR #223, opened 2020-08-18
#222 - Added Authentication.Joined

### LINKED ISSUE #222: Authentication composition
Multiple `Authentication` instances might be joined into single `Authentication` that delegates to others. In production it might be composition of different authentication methods. It is also useful in tests to join several `Authentication.Single` instances to have `Authentication` allowing more then one user

olegmoz:
Closes #222 
Added `Authentication.Joined` class

### DIFF
--- a/src/main/java/com/artipie/http/auth/Authentication.java
+++ b/src/main/java/com/artipie/http/auth/Authentication.java
@@ -23,7 +23,10 @@
  */
 package com.artipie.http.auth;
 
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 /**
  * Authentication mechanism to verify user.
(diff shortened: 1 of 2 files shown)


src/main/java/com/artipie/http/auth/Authentication.java
```diff
@@ -101,4 +104,44 @@ public Single(final String username, final String password) {
...
+        public Joined(final Authentication... origins) {
+            this(Arrays.asList(origins));
+        }
+
+        /**
+         * Ctor.
+         *
+         * @param origins Origin authentications.
+         */
+        public Joined(final Collection<Authentication> origins) {
```
g4s8, line 135: @olegmoz `Collection` type seems to be not correct here, since it can be unordered (e.g. `Set<>`) but it looks logical to check each authentication in user-specified order, so I'd use `List<>` here instead of `Collection<>`

olegmoz:
@g4s8 thanks, I've changed `Collection` to `List` and copied `toString` from `ChainedAuth`. Issue for adopting this new class in Artipie has been added as well, see https://github.com/artipie/artipie/issues/454

0crat:
Job `gh:artipie/http#223` is not assigned, can't get performer

Issue #238, opened 2020-09-14
Create interface for permission action


olenagerasimova:
Part of https://github.com/artipie/artipie/issues/574:
1. Create interface for permission action and synonyms, implementations for permissions `READ` and `WRITE`, create class to obtain this interface instance by any string synonym
2. Change `Permission.ByName` string action field to interface and add new ctor, use a set of synonyms to call `Permissions#allowed` with from the `Permission.ByName#allowed`

0crat:
@olegmoz[/z](https://www.0crat.com/u/olegmoz) the issue is closed not by @olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) (its creator); I won't close the order; please, re-open it and ask @olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) to close it

0crat:
@olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) resigned from [#238](https://github.com/artipie/http/issues/238), since the job is not in scope anymore

PR #239, opened 2020-09-14
#238 - created `Action` interface and its implementation

olenagerasimova:
Part of #238 
Created `Action` interface and its enum implementation.

### DIFF
--- a/src/main/java/com/artipie/http/auth/Action.java
+++ b/src/main/java/com/artipie/http/auth/Action.java
@@ -1,0 +1,115 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2020 Artipie
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


src/main/java/com/artipie/http/auth/Action.java
```diff
@@ -0,0 +1,115 @@
...
+ * Artipie permission action.
+ * @since 0.16
+ */
+public interface Action {
+
+    /**
+     * Action synonyms collection.
+     * @return Collection of synonyms
+     */
+    Collection<String> synonyms();
```
olegmoz, line 40: @olenagerasimova using word `synonym` implies that we have some main name, which is not the case. I'd suggest to rename it to `names`

src/main/java/com/artipie/http/auth/Action.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Action enumeration.
+     * @since 0.16
+     */
+    enum Actions implements Action {
```
olegmoz, line 46: @olenagerasimova with this nested enum we are going to access standard actions with expression like `Action.Actions.READ`. Does not look great. It will make more sense to rename this enum to `Standard` or maybe it will be even better to change `READ` and `WRITE` to be `public static` fields in `Action` interface. This way we are going to access standard actions with `Action.Standard.XYZ` or `Action.XYZ`.

olegmoz (changes requested):
@olenagerasimova thank you, please see two suggestions about naming

src/main/java/com/artipie/http/auth/Action.java
olenagerasimova: @olegmoz I'd like to keep enum as it allows iterating over items, creating `ByString` class for obtaining enum item from string. I'll use this class later in `Permission.ByName` to keep ctor from string action representation.

olenagerasimova:
@olegmoz thanks, grate naming suggestions, corrected and answered

src/main/java/com/artipie/http/auth/Action.java
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Artipie permission action.
+ * @since 0.16
+ */
+public interface Action {
+
+    /**
+     * Action synonyms collection.
+     * @return Collection of synonyms
+     */
```
olegmoz, line 39: @olenagerasimova please update method description to reflect recent changes

olegmoz (changes requested):
@olenagerasimova thanks! Please update JavaDocs as well

olenagerasimova:
@olegmoz thanks, corrected

olegmoz (approved):

0crat:
Job `gh:artipie/http#239` is not assigned, can't get performer

PR #245, opened 2020-09-25
#244 - work with Authentication.User instead of username

### LINKED ISSUE #244: Change auth interfaces signature
Change signature of `Identities`, `Permission` and `Permissions` interfaces to work with `Authentication.User` class interad of just user name.

olenagerasimova:
Closes #244 
Made `Identities`, `Permission` and `Permissions` work with `Authentication.User` instead of user name.

### DIFF
--- a/src/main/java/com/artipie/http/auth/Authentication.java
+++ b/src/main/java/com/artipie/http/auth/Authentication.java
@@ -39,11 +39,6 @@ import java.util.stream.Stream;
 public interface Authentication {
 
     /**
-     * Super user instance.
-     */
-    User SUPER_USER = new User("*");
-
-    /**
      * Find user by credentials.
      * @param username Username
      * @param password Password
--- a/src/main/java/com/artipie/http/auth/Authentication.java
+++ b/src/main/java/com/artipie/http/auth/Authentication.java
@@ -39,6 +39,11 @@ import java.util.stream.Stream;
 public interface Authentication {
 
     /**
+     * Super user instance.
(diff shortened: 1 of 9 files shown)


src/main/java/com/artipie/http/auth/Authentication.java
```diff
@@ -38,6 +38,11 @@
  */
 public interface Authentication {
 
+    /**
+     * Super user instance.
+     */
+    User SUPER_USER = new User("*");
```
olegmoz, line 44: @olenagerasimova meaning of "super" user is unclear. It seams that it's purpose is something like "any user" and is used specifically to check permissions for any user in `Permissions`. So I'd suggest to change the name and move the constant to `Permissions`

src/main/java/com/artipie/http/auth/Permissions.java
```diff
@@ -37,11 +37,11 @@
 
     /**
      * Check if user allowed to perform action.
-     * @param name User name
+     * @param user User name
```
olegmoz, line 40: @olenagerasimova comment for the argument needs an update too

olegmoz (changes requested):
@olenagerasimova thanks, please see two comments

olenagerasimova:
@olegmoz thanks, all fixed

olegmoz (approved):

0crat:
Job `gh:artipie/http#245` is not assigned, can't get performer

PR #248, opened 2020-10-02
#247 - created BasicAuthSlice

### LINKED ISSUE #247: Creare `BasicAuthSlice`
Let's create `BasicAuthSlice` to john `SliceAuth` and `BasicIdentities` functionality and deprecate `SliceAuth`, `Identities` and `BasicIdentities`.

olenagerasimova:
Closes #247 
I've created `BasicAuthSlice` from `SliceAuth` and `BasicIdentities` and deprecated `Identities`, `SliceAuth` and `BasicIdentities`

### DIFF
--- a/src/main/java/com/artipie/http/auth/BasicAuthSlice.java
+++ b/src/main/java/com/artipie/http/auth/BasicAuthSlice.java
@@ -26,65 +26,60 @@ package com.artipie.http.auth;
 import com.artipie.http.Headers;
 import com.artipie.http.Response;
 import com.artipie.http.Slice;
+import com.artipie.http.headers.Authorization;
 import com.artipie.http.headers.WwwAuthenticate;
+import com.artipie.http.rq.RqHeaders;
 import com.artipie.http.rs.RsStatus;
 import com.artipie.http.rs.RsWithHeaders;
 import com.artipie.http.rs.RsWithStatus;
 import java.nio.ByteBuffer;
 import java.util.Map;
+import java.util.Optional;
+import org.cactoos.text.Base64Decoded;
 import org.reactivestreams.Publisher;
 
 /**
- * Slice wrapper with authorization and authentication.
- * <p>
- * Example: the class which allows upload only for users with 'upload' permissions
- * and resolves user identity by "http-basic" mechanism:
- * </p>
- * <pre><code>
- * new SliceAuth(
- *   new SliceUpload(storage),
- *   new Permission.ByName("upload", permissions),
- *   new AuthBasic(passwords)
- * );
- * </code></pre>
- * @since 0.8
- * @deprecated Use {@link BasicAuthSlice} instead
+ * Slice with basic authentication.
+ * @since 0.17
  */
-@Deprecated
-public final class SliceAuth implements Slice {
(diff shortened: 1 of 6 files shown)


src/main/java/com/artipie/http/auth/BasicAuthSlice.java
```diff
@@ -0,0 +1,122 @@
...
+                }
+            ).orElseGet(
+                () -> {
+                    final Response rsp;
+                    if (this.perm.allowed(Permissions.ANY_USER)) {
+                        rsp = this.origin.response(line, headers, body);
+                    } else {
+                        rsp = new RsWithHeaders(
+                            new RsWithStatus(RsStatus.UNAUTHORIZED),
+                            new Headers.From(new WwwAuthenticate("Basic"))
```
olegmoz, line 101: @olenagerasimova we have `PREFIX` constant that is also `"Basic"`, but with space. Also `Basic` auth scheme name is used in `Authorization.Basic` class. I'd propose to join them all into single constant.

src/test/java/com/artipie/http/auth/BasicAuthSliceTest.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Test
+    void returnsForbiddenIfNotAllowed() {
+        final String name = "john";
+        MatcherAssert.assertThat(
+            new BasicAuthSlice(
+                new SliceSimple(new RsWithStatus(RsStatus.OK)),
+                (user, pswd) -> Optional.of(new Authentication.User(name)),
+                new Permission.ByName("read", (user, action) -> false)
```
olegmoz, line 90: @olenagerasimova I think we can simplify `new Permission.ByName("read", (user, action) -> false)` to just `(user, action) -> false`

olegmoz (changes requested):
@olenagerasimova thanks, please see two comments

olenagerasimova:
@olegmoz thanks, corrected. Also, I've added one more test case to make sure headers are parsed correctly.

src/main/java/com/artipie/http/auth/BasicAuthSlice.java
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Slice with basic authentication.
+ * @since 0.17
+ */
+public final class BasicAuthSlice implements Slice {
+
+    /**
+     * Basic authentication prefix.
+     */
+    public static final String PREFIX = "Basic";
```
olegmoz, line 50: @olenagerasimova name of this field is not really appropriate especially when exposed outside of class. It is a "prefix" when you look at header value as a string, but it's meaning is name authentication scheme which could also be `Bearer`, `Digest` etc. I'd suggest to change this field name to `SCHEME`, that would make more sense to one who reads the code.

olegmoz (changes requested):
@olenagerasimova thanks, still have one concern, please take a look

olenagerasimova:
@olegmoz thanks, renamed

olegmoz (approved):

0crat:
Job `gh:artipie/http#248` is not assigned, can't get performer

PR #252, opened 2020-10-14
#251 - StatefulResponse renamed to CachedResponse and fixed

### LINKED ISSUE #251: StatefulResponse finishes send before body is read
`StatefulResponse` returns completed future from it's `send()` method before it reads body. By contract body publisher not guaranteed to be readable after send is finished, because sending reply may wish to dispose allocated resources. `StatefulResponse` should be fixed to be compatible with all responses, including responses producing one-time content publishers such as HTTP client responses.

olegmoz:
Closes #251 
`StatefulResponse` renamed to `CachedResponse` and changed to keep copy of body


olenagerasimova (approved):
@olegmoz thanks, looks good to merge

0crat:
Job `gh:artipie/http#252` is not assigned, can't get performer

Issue #263, opened 2020-11-24
`BasicAuthSlice` should return forbidden when permissions prohibit


genryxy:
`BasicAuthSlice` should return `FORBIDDEN` when permissions prohibit. Not it returns `OK` when Basic authorization header is correct but permission forbid.

genryxy:
`BasicAuthSlice` returns `OK` when permissions prohibit but each user is allowed to perform all actions from yaml settings file. So, the current behavior is correct therefore close this issue.

PR #264, opened 2020-11-24
#263 - Add disabled test for BasicAuthSlice

genryxy:
Part of #263 
Add disabled test for `BasicAuthSlice` to verify that response returns `FORBIDDEN` when permissions prohibit.


olegmoz:
@genryxy I have tried to upgrade HTTP to `0.16.9` in artipie/artipie and everything works fine in this case. So I believe there is no problem we recent changes to `BasicAuthSlice` and the test case in this PR works as expected.

Issue #261, opened 2020-11-20
Minimal support for Bearer authentication


olegmoz:
Add Bearer authentication support. The module should provide `Slice` implementation that may wrap another `Slice` and add bearer authentication functionality for it. Bearer authentication is described in RFC: https://tools.ietf.org/html/rfc6750

olegmoz:
@rultor release, tag=`0.17`

PR #267, opened 2020-11-26
#261 - Added challenge to BearerAuthScheme

olegmoz:
Part of #261 
Added challenge to `BearerAuthScheme`


olenagerasimova (approved):
@olegmoz looks good to me

PR #272, opened 2020-11-27
#261 - Added absent auth scheme

olegmoz:
#261 
Added absent auth scheme that is useful as `AuthScheme` implementation when no auth is required.


olenagerasimova (approved):
@olegmoz thanks, looks good to me

olegmoz:
@rultor release, tag=`0.17.1`

PR #281, opened 2020-12-20
Multipart implementation

### LINKED ISSUE #90: Multipart.java:48-51: Finish implementation. In order to...
The puzzle `32-8c776bfe` from #32 has to be resolved: 

https://github.com/artipie/http/blob/16bfad340ea584590ad982844710bdc96e398f27/src/main/java/com/artipie/http/rq/Multipart.java#L48-L51

The puzzle was created by @Sammers21 on 19-Mar-20. 

Estimate: 60 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

Sammers21:
Closes #90

### DIFF
--- a/src/main/java/com/artipie/http/rq/Multipart.java
+++ b/src/main/java/com/artipie/http/rq/Multipart.java
@@ -24,17 +24,22 @@
 
 package com.artipie.http.rq;
 
-import com.artipie.http.stream.ByteByByteSplit;
+import com.artipie.http.headers.Header;
 import io.reactivex.Flowable;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 import org.reactivestreams.Processor;
-import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
 
@@ -43,12 +48,9 @@ import org.reactivestreams.Subscription;
  * See
  * <a href="https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html">rfc1341</a>
  * spec.
- * @todo #32:60min Finish implementation.
- *  In order to ensure that multipart parser works correctly, the MultipartTest has been written.
- *  The test is disabled for now, but, when this class will be fully implemented, the test should be
- *  enable.
  * @since 0.4
  * @checkstyle ConstantUsageCheck (500 lines)
+ * @checkstyle StringLiteralsConcatenationCheck (500 lines)
  */
 @SuppressWarnings({"PMD.AvoidDuplicateLiterals", "PMD.UnusedPrivateField", "PMD.SingularField"})
 public final class Multipart implements Processor<ByteBuffer, Part> {
@@ -59,14 +61,24 @@ public final class Multipart implements Processor<ByteBuffer, Part> {
     private static final String CRLF = "\r\n";
 
     /**
-     * The subscriber part.
+     * The subscriber.
      */
-    private final Subscriber<ByteBuffer> subscriber;
+    private final AtomicReference<Subscriber<? super Part>> dsub;
 
     /**
-     * The publisher part.
+     * Is upstream publisher completed?
      */
-    private final Publisher<Part> publisher;
+    private final AtomicBoolean completed;
+
+    /**
+     * The buffers.
+     */
+    private final List<ByteBuffer> buffers;
+
+    /**
+     * The multipart boundary.
+     */
+    private String boundary;
 
     /**
      * Ctor.
@@ -74,7 +86,7 @@ public final class Multipart implements Processor<ByteBuffer, Part> {
      * @param headers Request headers.
      */
     public Multipart(final Iterable<Map.Entry<String, String>> headers) {
-        this(() -> boundary(headers));
+        this(() -> boundaryFromHeaders(headers));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Sammers21:
@olegmoz, review the PR please

src/main/java/com/artipie/http/rq/Multipart.java
```diff
@@ -43,12 +48,9 @@
  * See
  * <a href="https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html">rfc1341</a>
  * spec.
- * @todo #32:60min Finish implementation.
- *  In order to ensure that multipart parser works correctly, the MultipartTest has been written.
- *  The test is disabled for now, but, when this class will be fully implemented, the test should be
- *  enable.
  * @since 0.4
  * @checkstyle ConstantUsageCheck (500 lines)
+ * @checkstyle StringLiteralsConcatenationCheck (500 lines)
```
olegmoz, line 53: @Sammers21 do not see good reason to use string literals concatenation in this class

src/main/java/com/artipie/http/rq/Multipart.java
```diff
@@ -59,22 +61,32 @@
...
-     * The publisher part.
+     * Is upstream publisher completed?
      */
-    private final Publisher<Part> publisher;
+    private final AtomicBoolean completed;
+
+    /**
+     * The buffers.
+     */
+    private final List<ByteBuffer> buffers;
```
olegmoz, line 76: @Sammers21 as I see, the class just stores all buffers and does not really process them in reactive manner. This makes it pointless

src/main/java/com/artipie/http/rq/Multipart.java
```diff
@@ -91,41 +103,82 @@ public Multipart(final Supplier<String> boundary) {
...
+        if (!this.dsub.compareAndSet(null, sub)) {
+            throw new IllegalStateException("Only one subscription is allowed");
+        }
+        this.justParse();
     }
 
     @Override
     public void onSubscribe(final Subscription subscription) {
-        this.subscriber.onSubscribe(subscription);
+        subscription.request(Long.MAX_VALUE);
```
olegmoz, line 122: @Sammers21 I think that lack of backpressure support should be documented in class description

src/main/java/com/artipie/http/rq/Multipart.java
```diff
@@ -91,41 +103,82 @@ public Multipart(final Supplier<String> boundary) {
...
     @Override
     public void onNext(final ByteBuffer item) {
-        this.subscriber.onNext(item);
+        this.buffers.add(item);
     }
 
     @Override
     public void onError(final Throwable throwable) {
-        this.subscriber.onError(throwable);
+        throw new IllegalStateException(throwable);
```
olegmoz, line 132: @Sammers21 I believe the error should be propagated to subscription, jut throwing it here is pointless

src/main/java/com/artipie/http/rq/Multipart.java
```diff
@@ -91,41 +103,82 @@ public Multipart(final Supplier<String> boundary) {
...
+                            final String body = hnb[1];
+                            return new PartFromPublisher(
+                                headers,
+                                Flowable.just(
+                                    ByteBuffer.wrap(body.getBytes(StandardCharsets.UTF_8))
+                                )
+                            );
+                        })
+                .subscribe(sub);
+        }
```
olegmoz, line 181: @Sammers21 this method is too big, please split it into multiple methods

src/test/java/com/artipie/http/rq/MultipartTest.java
```diff
@@ -55,24 +54,28 @@
...
+                                Flowable.fromPublisher(mpp).flatMap(part -> part)
+                            ).value(),
+                            new IsEqual<>("Hello worrrrld!!!\r\nHello worrrrld!!!\r\n")
+                        );
+                        connection.accept(
+                            RsStatus.OK,
+                            Headers.EMPTY,
+                            Flowable.empty()
+                        );
+                    }).start();
```
olegmoz, line 78: @Sammers21 could you please explain why this code is running in separate `Thread`? Response send should be completed async when `connection.accept` is complete. Right now test looks to be broken. I am not sure that it really checks anything

olegmoz (changes requested):
@Sammers21 thanks, please see comments

src/main/java/com/artipie/http/rq/Multipart.java
Sammers21: @olegmoz, well, it's not pointless, it makes sense as a first step. The next step is to use ByteSplitter and make it fully reactive.

src/test/java/com/artipie/http/rq/MultipartTest.java
Sammers21: @olegmoz, because `new ByteFlowAsString(Flowable.fromPublisher(mpp).flatMap(part -> part)).value()` will cause a deadlock due to vert.x threading policy

src/main/java/com/artipie/http/rq/Multipart.java
Sammers21: @olegmoz,  makes sense.

src/main/java/com/artipie/http/rq/Multipart.java
olegmoz: @Sammers21 it could have been useful if we didn't have non-reactive way to parse multipart. But there is no rush to add some multipart support, we need reactive one. So I'd prefer not to merge non-reactive implementation, as it will give false impression of complete feature

src/test/java/com/artipie/http/rq/MultipartTest.java
olegmoz: @Sammers21 does not sound great. Cannot be sure it won't cause deadlock without a thread. The tests needs to be modified so it does not require to spawn a thread for the request

Sammers21: @olegmoz, Олег, что значит не уверен? Я тебе говорю 100% что будет деделок. Вместо того чтобы это писать, можно пойти и самому убедиться? Какой ужас

0crat:
Job `gh:artipie/http#281` is not assigned, can't get performer

PR #284, opened 2021-01-19
#283 - Obtain random free port

### LINKED ISSUE #283: Add class for obtaining free port
Let's introduce class `RandomFreePort` to get free port since this is required in many modules.

genryxy:
Closes #283 
Added class for obtaining random free port.

### DIFF
--- a/src/main/java/com/artipie/http/misc/RandomFreePort.java
+++ b/src/main/java/com/artipie/http/misc/RandomFreePort.java
@@ -30,33 +30,14 @@ import java.net.ServerSocket;
 /**
  * Provides random free port.
  * @since 0.18
+ * @checkstyle NonStaticMethodCheck (500 lines)
  */
 public final class RandomFreePort {
-    /**
-     * Random free port.
-     */
-    private final int port;
-
-    /**
-     * Ctor.
-     */
-    public RandomFreePort() {
-        this.port = RandomFreePort.freePort();
-    }
-
-    /**
-     * Returns free port.
-     * @return Free port
-     */
-    public int value() {
-        return this.port;
-    }
-
     /**
      * Returns free port.
      * @return Free port.
      */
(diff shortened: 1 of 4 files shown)


src/main/java/com/artipie/http/misc/RandomFreePort.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Random free port.
+     */
+    private final int port;
+
+    /**
+     * Ctor.
+     */
+    public RandomFreePort() {
+        this.port = RandomFreePort.freePort();
```
olenagerasimova, line 44: @genryxy hm, I do not think this class should have any fields, it does not make any sence, why we should obtain the port in the ctor. I'd suggest to remove the field and put all logic to the `value()` method and may be rename the `value()` method into `get()` or `obtain()`

olenagerasimova (changes requested):
@genryxy please, check my comment

src/main/java/com/artipie/http/misc/RandomFreePort.java
genryxy: @olenagerasimova but in this case method `value()` would be `public static` that is not good. Or it would be better to get rid of field and rework method `value()`?

olenagerasimova: @genryxy why should it be `static`? I'd suggest to suppress corresponding checkstyle error, do not make the method static, and then usage of this class will look like `int port = new RandomFreePort().get()`

genryxy:
@olenagerasimova thanks, corrected

genryxy:
@olegmoz could you please merge this PR

0crat:
Job `gh:artipie/http#284` is not assigned, can't get performer

PR #296, opened 2021-04-20
Pass query params from servlet request to slice

### LINKED ISSUE #295: Servlet slise wrapper doesn't parse URI query params
The slice under servlet wrapper doesn't receive query params from URI.

g4s8:
Fixes #295 - reproduced with a test by passing query params to
request URI, and fixed by passing query params to slice from servlet
wrapper. Found and fixed a bug with query params decoding,
added more tests for slice wrapper.


olenagerasimova (approved):
@g4s8 thanks

PR #298, opened 2021-04-23
Added class to parse `Accept` header

### LINKED ISSUE #297: `Accept` header
Let's add class to parse `Accept` header values according to documentation
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

olenagerasimova:
Closes #297 
Added class to parse `Accept` header and corresponding test.

### DIFF
--- a/src/main/java/com/artipie/http/headers/Accept.java
+++ b/src/main/java/com/artipie/http/headers/Accept.java
@@ -89,7 +89,7 @@ public final class Accept {
             }
         );
         return map.entrySet().stream()
-            .sorted((one, two) -> Float.compare(one.getValue(), two.getValue()) * (-1))
+            .sorted((one, two) -> Float.compare(two.getValue(), one.getValue()))
             .map(Map.Entry::getKey).collect(Collectors.toList());
     }
 }
--- a/src/main/java/com/artipie/http/headers/Accept.java
+++ b/src/main/java/com/artipie/http/headers/Accept.java
@@ -1,0 +1,84 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2020 Artipie
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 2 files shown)


src/main/java/com/artipie/http/headers/Accept.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Headers.
+     */
+    private final Iterable<Map.Entry<String, String>> headers;
+
+    /**
+     * Ctor.
+     * @param headers Headers to extract `accept` header from
+     */
+    public Accept(final Iterable<Map.Entry<String, String>> headers) {
```
genryxy, line 55: @olenagerasimova probably pass `Headers` instead of `Iterable<Map.Entry<String, String>>` as it `Headers` extends the last one

src/test/java/com/artipie/http/headers/AcceptTest.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+class AcceptTest {
+
+    @Test
+    void parsesAndSortsHeaderValues() {
+        MatcherAssert.assertThat(
+            new Accept(
+                new Headers.From(
+                    new Header("Accept", "text/html, application/xml;q=0.9, audio/aac;q=0.4"),
```
genryxy, line 43: @olenagerasimova is it possible that the first header and the second will contain the same value, for example `text/html`

genryxy (changes requested):
@olenagerasimova thanks, two comments

src/main/java/com/artipie/http/headers/Accept.java
olenagerasimova: @genryxy hm, I'd prefer to leave `Iterable` as it's more general and we can pass `Headers`

olenagerasimova:
@genryxy thanks, fixed and answered

genryxy (dismissed):
@olenagerasimova thanks, looks good to me

src/main/java/com/artipie/http/headers/Accept.java
```diff
@@ -0,0 +1,95 @@
...
+                        if (val != null) {
+                            min = Float.min(weight, val);
+                        }
+                        return min;
+                    }
+                );
+            }
+        );
+        return map.entrySet().stream()
+            .sorted((one, two) -> Float.compare(one.getValue(), two.getValue()) * (-1))
```
g4s8, line 92: @olenagerasimova you could just switch arguments of `Float.compare` instead of `* (-1)`:
```java
.sorted((one, two) -> Float.compare(two.getValue(),  one.getValue()))
```

g4s8 (changes requested):
@olenagerasimova just a minor comment

olenagerasimova:
@g4s8 thanks, done

g4s8 (approved):
@olenagerasimova thanks

Issue #299, opened 2021-05-14
Reactive multipart request


g4s8:
Add reactive multipart request's wrapper. It should support [RFC1341-7.2](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html) request message format. The class should accept `Request` interface as constructor param and provide a method `Publisher<Request> parts()` with body parts of multipart request.

Impl. notes:
 - Since the multipart body is always received sequentially part by part, the `parts()` method should not publish the next part until the previous is fully read.
 - The implementation should not keep request part data in memory or storage, it should process each chunk and send to proper downstream.

0pdd:
@g4s8 the puzzle [#334](https://github.com/artipie/http/issues/334) is still not solved.

PR #300, opened 2021-05-14
Classes with stubs for multipart request and test

### LINKED ISSUE #32: Reactive parser for multipart.

### LINKED ISSUE #90: Multipart.java:48-51: Finish implementation. In order to...

g4s8:
For #299 - added `RqMultipart` class for multipart request parsing
and `Part` nested class as part publisher implementation. Added
unit test for parsing canonical example from RFC specification.

Closes #32 and closes #90 as outdated and no longer valid.


src/test/java/com/artipie/http/rq/RqMultipartTest.java
```diff
@@ -0,0 +1,89 @@
...
+            new RqMultipart(
+                new ContentType("multipart/mixed; boundary=\"simple boundary\""),
+                new Content.From(simple.getBytes(StandardCharsets.US_ASCII))
+            ).parts()
+        ).<String>flatMapSingle(
+            part -> Single.just(part).to(ContentAs.STRING)
+        ).toList().blockingGet();
+        MatcherAssert.assertThat(
+            parsed,
+            Matchers.contains(first, second)
```
olenagerasimova, line 86: @g4s8 I'd suggest to use more strict matcher here, `contains` seems too weak, result can contain whatever else

olenagerasimova (changes requested):
@g4s8 thanks, check my suggestion please

olenagerasimova (approved):

PR #307, opened 2021-05-20
ByteBuffer chunks tokenizer with a test

g4s8:
For #299 - added tokenizer for `ByteBuffer` chunks processing to split
by delimiter into series of chunks sequences. Added tests with different
scenarios. Moved all multipart related classes to a new package. Removed
unfinished implementation with similar purpose, due to incorrect design.

(this PR has checkstyle & PMD suppressions due to high performance requirements and low-level operations with a `ByteBuffer`; due to increased complexity it has inline comments)


src/test/java/com/artipie/http/rq/multipart/ByteBufferTokenizerTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Test case for ByteBufferTokenizer.
+ *
+ * @since 1.0
+ * @checkstyle ParameterNumberCheck (500 lines)
+ */
+@SuppressWarnings("PMD.UseObjectForClearerAPI")
+final class ByteBufferTokenizerTest {
+
```
genryxy, line 26: @g4s8 thanks, probably also to add test when length some of the parts of splitted source string is more then capacity of the buffer

g4s8: @genryxy sure, it's done on 39 line of the test: here I added a for loop to try each stream with a capacity from 1 to full string length

genryxy (approved):

PR #316, opened 2021-05-24
Added `override` parameter to `RsWithHeaders`

### LINKED ISSUE #305: `RsWithBody` duplicates content-length header
`RsWithBody`should add content-length header if this header is absent, do not duplicate it.

olenagerasimova:
Closes #305 
Added `override` parameter to `RsWithHeaders`: if this parameter is true, origin header values are replaced with provided headers values. This feature is now used in `RsWithBody` to avoid `Content-Length` header duplication/mismatch.

### DIFF
--- a/src/main/java/com/artipie/http/rs/RsWithHeaders.java
+++ b/src/main/java/com/artipie/http/rs/RsWithHeaders.java
@@ -146,7 +146,8 @@ public final class RsWithHeaders implements Response {
                 this.headers.forEach(list::add);
                 hrs.forEach(
                     item -> {
-                        if (list.stream().noneMatch(val -> val.getKey().equals(item.getKey()))) {
+                        if (list.stream()
+                            .noneMatch(val -> val.getKey().equalsIgnoreCase(item.getKey()))) {
                             list.add(item);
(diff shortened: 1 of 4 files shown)


src/main/java/com/artipie/http/rs/RsWithHeaders.java
```diff
@@ -113,11 +140,22 @@ private ConWithHeaders(
...
-                new Headers.From(this.headers, hrs),
-                body
-            );
+            final Headers res;
+            if (this.override) {
+                final List<Map.Entry<String, String>> list = new ArrayList<>(10);
+                this.headers.forEach(list::add);
+                hrs.forEach(
+                    item -> {
+                        if (list.stream().noneMatch(val -> val.getKey().equals(item.getKey()))) {
```
g4s8, line 149: According to HTTP spec the header name is case-insensitive, so we need to compare header names ignoring case

g4s8 (changes requested):
@olenagerasimova just one comment

olenagerasimova:
@g4s8 thanks, corrected

PR #318, opened 2021-05-24
Added multipart headers builder

g4s8:
For #229 - added classes to build multipart request headers from ByteBuffer chunks.


src/main/java/com/artipie/http/rq/multipart/MultipartHeaders.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Push new chunk to builder.
+     * @param chunk Part of headers bytes
+     */
+    void push(final ByteBuffer chunk) {
+        if (this.data == null) {
+            throw new IllegalStateException("Headers were constructured");
+        }
+        synchronized (this.lock) {
```
olenagerasimova, line 90: @g4s8 seems like another check `this.data == null` is required after the lock was obtained

olenagerasimova (changes requested):
@g4s8 thanks, hm, what if we use `Optional` for `data` and `cache` to avoid `nulls`? And there is another comment

g4s8:
@olenagerasimova thanks, the `null` is used here only to notify the GC that the buffer and it's back-end byte-array can be removed from heap. I moved the common logic of working with byte-buffers into a new class, please see new commit: https://github.com/artipie/http/pull/318/commits/33c0d78ddca896648cf0ad9fddd1f36d982222b3

olenagerasimova (approved):

Issue #324, opened 2021-06-04
Response matchers


g4s8:
Add more response matchers

g4s8:
@artipie/members do you have a need for some additional matchers for http response?

olenagerasimova:
@g4s8 It would be convenient to have a response matcher for string, so we could use `StringContains()` etc

PR #325, opened 2021-06-04
Response matcher for JSON

g4s8:
For #324 - added JSON response matcher accepting
`? extends JsonStructure` - it's either `JsonObject` or
`JsonArray`. Added tests for matcher.


genryxy (approved):
@g4s8 thanks, approve

PR #333, opened 2021-06-10
Added reactive processor tokenizer

g4s8:
For #299 - reactive processor implementation
based on ByteBufferTokenizer - it takes ByteBuffer
publisher as upstream and split it by delimiters into
ByteBuffers and send to downstream.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -235,7 +235,7 @@ SOFTWARE.
                     <limit>
                       <counter>CLASS</counter>
                       <value>MISSEDCOUNT</value>
-                      <maximum>25</maximum>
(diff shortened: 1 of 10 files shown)


g4s8:
@genryxy you may want to use this class for Helm metadata processing too

olenagerasimova (approved):
@g4s8 thanks

genryxy:
@g4s8 could you please release a new version, so I would be able to use `TokenizerFlatProc` in Helm adapter

g4s8:
@genryxy done: https://github.com/artipie/http/issues/344

PR #340, opened 2021-06-29
Implemented multipart request

g4s8:
Finished multipart request parser as `RqMultipart` class,
added test to verify against full multipart request.
    
Closes: #299


olenagerasimova (approved):
@g4s8 thanks

PR #355, opened 2021-08-16
Added IT for slices

olenagerasimova:
Part of #356
Added IT to illustrate the situation when Artipie stacks on the second request. Note, that slice should be introduced over `SliceRoute` and `RtRulePath` chain as `SliceSimple` with some response. If `SliceSimple` is replaced with a proper separate `Slice` implementation (as [AuthTokenSlice.java](https://github.com/artipie/conda-adapter/blob/master/src/main/java/com/artipie/conda/http/AuthTypeSlice.java) in conde for example) everything works properly.

Enable and run disabled test to see the problem.


olenagerasimova:
@g4s8 please, could you take a look and help to understand where is the problem?

g4s8 (changes requested):
@olenagerasimova it looks like a bug, can you create a ticket please and describe this issue there?

olenagerasimova:
@g4s8 created #356

g4s8 (approved):
@olenagerasimova thanks

Issue #356, opened 2021-08-16
Artipie stacks on the second request


olenagerasimova:
When slice is introduced over `SliceRoute` and `RtRulePath` chain as `SliceSimple` with some response, for example
```java
    /**
     * Test slice implementation.
     * @since 0.20
     */
    private static final class TestSlice extends Slice.Wrap {

        /**
         * Ctor.
         */
        protected TestSlice() {
            super(
                new SliceRoute(
                    new RtRulePath(
                        new ByMethodsRule(RqMethod.GET),
                        new BasicAuthSlice(
                            new SliceSimple(
                                new RsJson(
                                    () -> Json.createObjectBuilder().add("any", "any").build()
                                )
                            ),
                            Authentication.ANONYMOUS,
                            new Permission.ByName(Permissions.FREE, Action.Standard.READ)
                        )
                    )
                )
            );
        }
    }
```
and served over `VertxSliceServer` on the second request response is never sent and connection stays open forever. Request should be performed from the local pc/docker container.

Check #355 for an example.

PR #359, opened 2021-08-18
fix: duplicate buffers of RsWithBody

g4s8:
Reproduced #356 issue in `RsWithBody` unit test: on first request
VertX server reads `ByteBuffer` content and changes `position` of
buffer from 0 to `limit`, so next attempts to read the same buffer
will read empty buffers. The fix is: call `duplicate()` of
`ByteBuffer` on sending `RsWithBody` content to not share buffer
properties (only share buffer data), so all `position` changes
in duplicate won't affect origin buffer properties.

Fix: #356


olenagerasimova (approved):
@g4s8 thanks for the fix and explanation

Issue #360, opened 2021-08-20
Accept header with multiple content-types is not parsed correctly


g4s8:
A client request from Docker client with accept header `application/vnd.oci.image.manifest.v1+json,application/vnd.docker.distribution.manifest.v2+json,application/vnd.docker.distribution.manifest.v1+json,application/vnd.docker.distribution.manifest.list.v2+json` is not parsed correctly into multiple mime-type values, it's parsed into single invalid mime-type.

Related issue: https://github.com/artipie/docker-adapter/issues/475

PR #361, opened 2021-08-20
fix: parse accept header correctly

g4s8:
Parse MIME types string of `Accept` header correctly:
 1. Do not require whitespace after delimiter, `,` is enough
 2. Sort types correctly: 1) by qualifier 2) by type alphabetically,
 ignoring wildcards `*` 3) by subtype, same as type

Fix: #360
Related: https://github.com/g4s8/mime/pull/28


olenagerasimova (approved):

Issue #371, opened 2021-09-09
Token based authentication


olenagerasimova:
Anaconda repository uses token based authentication mechanism, so we need to support such authentication. Repository [uses](https://docs.anaconda.com/anacondaorg/user-guide/tasks/work-with-accounts/#cloud-accounts-tokens) random alphanumeric string with one year expire period.

olenagerasimova:
@g4s8 could you please take a look? Any ideas for general implementation? We need to store the tokens somewhere (in storage more likely) and choose an algorithm to generate tokens.

g4s8:
@olenagerasimova I think we can change implementation of `com.artipie.http.auth.AuthScheme` which is accepting only request headers now to get authentication result. For Conda we should check request line too. So `AuthScheme` interface can accept not only headers but request line to get the user.

Then, we can implement `Authentication` interface for Conda which accepts only token and return user if found for token.

And finally, it could be joined by `AuthSlice` in Conda adapter.

---

As I see in Conda [documentation](https://docs.anaconda.com/anacondaorg/user-guide/tasks/work-with-accounts/#cloud-accounts-tokens), tokens are generated by some administrators there, so it's logically to store them in the same place as other authentication credentials - `_credentials.yaml`.

WDYT?

olenagerasimova:
@g4s8 thanks, sounds logical. The only part I cannot get is about storing the tokens: documentation shows anaconda interface where users (and administrators) can manage tokens, but creating token manually (by user itself or administrator) is not required, anaconda can (and so should artipie) generate user token automatically. 
We have access to `_credentials.yaml` only on artipie/artipie level now, how do we write new tokens there?

g4s8:
@olenagerasimova it's not clear, where exactly a new token should be created? On what level and what operation

olenagerasimova:
@g4s8 token should be created on repository server (anaconda/artipie) side, here is some [doc](https://api.anaconda.org/docs#/authentication/post_authentications) (not very informative) and logs:
```
[DEBUG] vert.x-eventloop-thread-1 com.artipie.conda.BodyLoggingSlice: >> POST /authentications HTTP_1_1

Host: host.testcontainers.internal:61246
User-Agent: Anaconda-Client/1.8.0 (+https://anaconda.org)
Accept-Encoding: gzip, deflate
Accept: application/json
Connection: keep-alive
x-binstar-api-version: 1.8.0
Content-Type: application/json
Content-Length: 245
[...]

PR #372, opened 2021-09-09
Added rq line to AuthScheme and token auth header

olenagerasimova:
Part of #371 
Added request line to `AuthScheme` interface and token authorization header.


src/test/java/com/artipie/http/headers/AuthorizationTokenTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.core.IsEqual;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests for {@link Authorization.Token}.
+ *
+ * @since 0.12
+ */
```
genryxy, line 15: @olenagerasimova thanks, looks good. Only here `@since` tag is incorrect

olenagerasimova:
@genryxy thanks, corrected

genryxy (approved):

olenagerasimova:
@g4s8 please, take a look/merge

src/main/java/com/artipie/http/auth/AuthScheme.java
```diff
@@ -37,9 +37,10 @@ public String challenge() {
      * Authenticate HTTP request by it's headers.
      *
      * @param headers Request headers.
+     * @param line Request line.
      * @return Authentication result.
      */
-    CompletionStage<Result> authenticate(Iterable<Map.Entry<String, String>> headers);
+    CompletionStage<Result> authenticate(Iterable<Map.Entry<String, String>> headers, String line);
```
g4s8, line 43: I think we need to add `default` method for backward compatibility, since this class is used most probably in all adapters, it can be:
```java
default CompletionStage<Result> authenticate(Iterable<Map.Entry<String, String>> headers) {
  return this.authenticate(headers, "");
}
```

g4s8 (changes requested):
@olenagerasimova please check my comment

olenagerasimova:
@g4s8 thanks, added

g4s8 (approved):
@olenagerasimova thanks

PR #377, opened 2021-09-16
MultipartHeaders: header values are not case-insensitive

### LINKED ISSUE #376: Header values are not case-insensitive
`MultipartHeaders` class makes header value lowercase while selecting headers by provided name. Such behavior is not correct, as header values are not case-insensitive, only names are. Do not forget to add corresponding test after fixing.

olenagerasimova:
Closes #376 
Fixed `MultipartHeaders` not to make header values lowercase, added corresponding test case. Also, added changelog to release as soon as this PR will be merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v0.23.1
+
+ - fix: MultipartHeaders: header values are not case-insensitive (#376) 
+
 ## v0.23
 
  - feat: added rq line to AuthScheme and token auth header (#372) 
--- a/src/main/java/com/artipie/http/rq/multipart/MultipartHeaders.java
+++ b/src/main/java/com/artipie/http/rq/multipart/MultipartHeaders.java
@@ -66,7 +66,7 @@ final class MultipartHeaders implements Headers {
                                 final String[] parts = line.split(":");
                                 return new Header(
                                     parts[0].trim().toLowerCase(Locale.US),
-                                    parts[1].trim().toLowerCase(Locale.US)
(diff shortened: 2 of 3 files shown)


genryxy:
@olenagerasimova but [here](https://developer.mozilla.org/ru/docs/Web/HTTP/Headers/Content-Disposition#%D0%BA%D0%B0%D0%BA_%D0%B7%D0%B0%D0%B3%D0%BE%D0%BB%D0%BE%D0%B2%D0%BE%D0%BA_%D0%B2_%D1%81%D0%BE%D1%81%D1%82%D0%B0%D0%B2%D0%BD%D0%BE%D0%BC_%D1%82%D0%B5%D0%BB%D0%B5) it is written that values are case insensitive

olenagerasimova:
@genryxy hm, I do not see corresponding statement by your link, but [this](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) doc does not say anything about values and `com.artipie.http.headers.Header#equals` method compares values as is. Also, actually, I did not describe the problem correctly: the question is not in case-insensitive or not case-insensitive, it's that we should not make values lowercase while parsing them, we should return them as is. 

In pypi adapter package names are sent in Content-Disposition header 
```
Content-Disposition: form-data; name="content"; filename="My-Test.txt"
```
of the part of multipart body and should be saved by name, provided as `filename` value, with capital letters if there are any.

genryxy (approved):
@olenagerasimova thanks, got it

PR #386, opened 2021-09-24
test(multipart): added test for conda payload

g4s8:
Added test with real conda payload, checked that
`RqMultipart` can read it fully.


olenagerasimova:
@g4s8 thanks, I've finally figured it out. I was trying to map part right after `RqMultipart.filter()` expecting only one part, that corresponds the condition, but it'necessary to filter empty (those which not correspond to condition) parts first:
```
        final Pair<Content, String> size = Flowable.fromPublisher(
            new RqMultipart(
                new ContentType("multipart/mixed; boundary=\"92fd51d48f874720a066238b824c0146\""),
                new Content.From(payload)
            ).filter(hdrs -> new ContentDisposition(hdrs).fieldName().equals("content"))
        )
            // filter filtered parts
            .filter(part -> part.headers().iterator().hasNext())
            // and then map
            .<Pair<Content, String>>map(
                part -> new ImmutablePair<>(
                    new Content.From(part),
                    new ContentDisposition(part.headers()).fileName()
                )
            ).blockingFirst();
``` 

Also, I noticed that `RqMultipart.filter()` stacks if payload does not have proper closing boundary.

olenagerasimova (approved):

PR #399, opened 2021-10-29
feat: Creates HeadSlice

### LINKED ISSUE #397: Move Slice for handling HEAD requests from files-adapter
Move `Slice` implementation for handling `HEAD` HTTP requests from artipie/files-adapter here, make it more flexible to add additional data to response headers which may be needed for concrete adapters implementation (e.g. maven-adapter appends checksums for HEAD requests of artifacts)

_Originally posted by @olenagerasimova in https://github.com/artipie/files-adapter/pull/79#discussion_r737157273_

baudoliver7:
Closes #397 

- Creates `HeadSlice`
- Introduces tests
- Introduces puzzle to use `HeadSlice` instead of `BlobMetadataSlice` in `artipie/files-adapter` (after releasing `artipie/http` project)

### DIFF
--- a/src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
+++ b/src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
@@ -4,7 +4,6 @@
  */
 package com.artipie.http.slice;
 
-import com.artipie.asto.Content;
 import com.artipie.asto.Key;
 import com.artipie.asto.Storage;
 import com.artipie.http.Headers;
@@ -15,15 +14,15 @@ import com.artipie.http.headers.ContentFileName;
 import com.artipie.http.headers.ContentLength;
 import com.artipie.http.rq.RequestLineFrom;
 import com.artipie.http.rs.RsFull;
-import com.artipie.http.rs.RsStatus;
 import com.artipie.http.rs.RsWithBody;
-import com.artipie.http.rs.StandardRs;
+import com.artipie.http.rs.RsWithHeaders;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Function;
+import com.artipie.http.rs.StandardRs;
 import org.reactivestreams.Publisher;
 
 /**
@@ -90,13 +89,12 @@ public final class BlobMetadataSlice implements Slice {
                                         result = this.storage.size(key)
                                             .thenApply(
                                                 size ->
-                                                    new RsFull(
-                                                        RsStatus.OK,
+                                                    new RsWithHeaders(
+                                                        StandardRs.OK,
                                                         new Headers.From(
                                                             new ContentFileName(uri),
                                                             new ContentLength(size)
-                                                        ),
-                                                        Content.EMPTY
+                                                        )
                                                     )
                                             );
                                     } else {
--- a/src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
+++ b/src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
@@ -1,0 +1,135 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package com.artipie.http.slice;
+
+import com.artipie.asto.Content;
+import com.artipie.asto.Key;
+import com.artipie.asto.Storage;
+import com.artipie.http.Headers;
+import com.artipie.http.Response;
+import com.artipie.http.Slice;
+import com.artipie.http.async.AsyncResponse;
+import com.artipie.http.headers.ContentFileName;
+import com.artipie.http.headers.ContentLength;
+import com.artipie.http.rq.RequestLineFrom;
+import com.artipie.http.rs.*;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+import org.reactivestreams.Publisher;
+
+/**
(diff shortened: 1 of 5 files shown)


olenagerasimova (approved):
@baudoliver7 thanks, looks good

src/test/java/com/artipie/http/slice/BlobMetadataSliceTest.java
```diff
@@ -0,0 +1,80 @@
...
+    private final Storage storage = new InMemoryStorage();
+
+    /**
+     * Tests that headers are correct when returning found.
+     *  @todo #397:30min Fix {@link RsFull} that duplicates header.
+     *   We notice (in this test) that due to the empty content, {@link RsFull}
+     *   firstly add a content length with value of 0. After that, BlobMetadataSlice
+     *   adds a content length with size of the blob found. We think that {@link RsFull}
+     *   should avoid header duplication. It should overwrite the existing one. After
+     *   fixing this issue, remove {@code new ContentLength(0)} in the test below.
```
g4s8, line 44: It's not a valid issue if you use `RsWithHeaders`

src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
```diff
@@ -0,0 +1,119 @@
...
+ * A {@link Slice} which only serves metadata on Binary files.
+ *
+ * @since 0.26.2
+ * @todo #397:30min Use this class in artipie/files-adapter.
+ *  We should replace {@link BlobMetadataSlice} of artipie/files-adapter by
+ *  this one. Before doing this task, we need to fix {@link RsFull} about header
+ *  duplication and the next release of artipie/http.
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
+ */
+public final class BlobMetadataSlice implements Slice {
```
g4s8, line 39: I think the name here should be more generic than for files-adapter, e.g. it could be `HeadSlice` or similar

src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
```diff
@@ -0,0 +1,119 @@
...
+                                    if (exist) {
+                                        result = this.storage.size(key)
+                                            .thenApply(
+                                                size ->
+                                                    new RsFull(
+                                                        RsStatus.OK,
+                                                        new Headers.From(
+                                                            new ContentFileName(uri),
+                                                            new ContentLength(size)
+                                                        ),
```
g4s8, line 98: This part should be moved to a separate function field, since we may have another headers to return in different adapters, e.g. maven adapter reads checksum files to add it to headers. The default behavior is OK now, so you can just move this implementation to default parameter in secondary constructor. See https://github.com/artipie/http/issues/397:
> make it more flexible to add additional data to response headers which may be needed for concrete adapters implementation

src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
```diff
@@ -0,0 +1,119 @@
...
+                        final Key key = this.transform.apply(uri.getPath());
+                        return this.storage.exists(key)
+                            .thenCompose(
+                                exist -> {
+                                    final CompletionStage<Response> result;
+                                    if (exist) {
+                                        result = this.storage.size(key)
+                                            .thenApply(
+                                                size ->
+                                                    new RsFull(
```
g4s8, line 93: You can use [RsWithHeaders](https://github.com/artipie/http/blob/master/src/main/java/com/artipie/http/rs/RsWithHeaders.java#L47) class to wrap `OK` response without using `RsFull` since we don't need body here:
```java
new RsWithHeaders(
  new Headers.From(...),
  StandardRs.OK
)
```
Then you won't get content-length duplication

src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+import org.reactivestreams.Publisher;
+
+/**
+ * A {@link Slice} which only serves metadata on Binary files.
+ *
+ * @since 0.26.2
+ * @todo #397:30min Use this class in artipie/files-adapter.
```
g4s8, line 33: And this puzzle is also not valid then

g4s8 (changes requested):
@baudoliver7 please check my comments

src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
baudoliver7: @g4s8 let's use `HeadSlice`. It's more convenient :)

src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
baudoliver7: @g4s8 regardless of adapter, we always have uri and length content. So wouldn't it be better to simply pass via constructor only additional headers if they exists ?

src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
baudoliver7: @g4s8 we need also release of artipie/http (that can take one day) as prerequisites before using it in `artipie/files-adapter` :)

src/main/java/com/artipie/http/slice/BlobMetadataSlice.java
baudoliver7: @g4s8 you will see in the latest commit my suggestion.

baudoliver7:
@g4s8 please, could you review my latest changes and my comments ?

src/main/java/com/artipie/http/slice/HeadSlice.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Path to key transformation.
+     */
+    private final Function<String, Key> transform;
+
+    /**
+     * Other headers specific to the storage.
+     */
+    private final Function<Storage, CompletionStage<Headers>> addhdrs;
```
g4s8, line 52: This is not correct:
 1. `Storage` is already encapsulated - no need to pass it as a parameter
 2. It's not possible to read request data in this function, which is required to determine head response

I think we can go with one of these options: either change it to `BiFunction<String, Headers, CompletionStage<Headers>>` to take request line and request headers as parameters and return async response headers; or create custom `@FunctionalInterface` for this type: `(String, Headers) -> CompletionStage<Headers>`

src/main/java/com/artipie/http/slice/HeadSlice.java
```diff
@@ -0,0 +1,134 @@
...
+            sto -> CompletableFuture.completedFuture(Headers.EMPTY)
+        );
+    }
+
+    /**
+     * Slice by key from storage using custom URI path transformation.
+     *
+     * @param storage Storage
+     * @param transform Transformation
+     * @param addhdrs Additional headers
```
g4s8, line 72: It's not actually additional headers but the function to make response headers from request parameters

src/main/java/com/artipie/http/slice/HeadSlice.java
```diff
@@ -0,0 +1,134 @@
...
+     */
+    public HeadSlice(
```
g4s8, line 74: We need a secondary constructor with default implementation of headers func as in current `response` method

src/main/java/com/artipie/http/slice/HeadSlice.java
```diff
@@ -0,0 +1,134 @@
...
+                                                        new ContentLength(size)
+                                                    )
+                                                )
+                                            )
+                                            .thenCompose(
+                                                res -> this.addhdrs.apply(this.storage)
+                                                    .thenApply(
+                                                        hdrs -> new RsWithHeaders(res, hdrs)
+                                                    )
+                                            );
```
g4s8, line 116: This should be the result of response headers func:
```java
this.resheaders.apply(line, headers).thenApply(headers -> new RsWithHeaders(StandardRs.OK, headers))
```
```

g4s8 (changes requested):
@baudoliver7 please check my comments

src/main/java/com/artipie/http/slice/HeadSlice.java
baudoliver7: @g4s8 The first suggestion is ok for me :)

src/main/java/com/artipie/http/slice/HeadSlice.java
baudoliver7: @g4s8 agree with you

src/main/java/com/artipie/http/slice/HeadSlice.java
baudoliver7: @g4s8 ok

g4s8 (approved):
@baudoliver7 thanks!

Issue #404, opened 2021-11-19
Implement GzipSlice


g4s8:
Add a new `Slice` implementation that checks that request `Accept-Encoding` header contains `gzip` value, e.g.: `Accept-Encoding: gzip` and compress output body with gzip and adds `Content-Encoding: gzip` header.

Headers docs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding
Required for: artipie/git-adapter#1

olenagerasimova:
@g4s8 it seems that it would be better to perform header checking on the upper level with the help of `RtRule.ByHeader`, not in the new `GzipSlice`. What do you say?

g4s8:
> 

@olenagerasimova `RtRule.ByHeader` is just a routing rule, but gzip compression here is a part of content-negotiation: if client support gzip content (send `accept-encoding` header), then server compress it; but if client doesn't support it, then server returns uncompressed response body.

olenagerasimova:
@g4s8 the content-negotiation logic can be handled with the help of the rules on the adapter level
```
new SliceRoute(
    RtRule.ByHeader("accept-encoding", ".*gzip.*"),
    new GzipSlicee()
),
new SliceRoute(
    RtRule.ByMethod(RqMethod.GET),
    new DownloadSlice()
),
```
Thus, `GzipSlice` will gzip the content and add `Content-Encoding: gzip` header, download slice will return content as is. Otherwise, we will duplicate the logic of `DownloadSlice` in the `GzipSlice`, which does not seem necessary.

g4s8:
> @g4s8 the content-negotiation logic can be handled with the help of the rules on the adapter level
> 
> ```
> new SliceRoute(
>     RtRule.ByHeader("accept-encoding", ".*gzip.*"),
>     new GzipSlicee()
> ),
> new SliceRoute(
>     RtRule.ByMethod(RqMethod.GET),
>     new DownloadSlice()
> ),
> ```
> 
> Thus, `GzipSlice` will gzip the content and add `Content-Encoding: gzip` header, download slice will return content as is. Otherwise, we will duplicate the logic of `DownloadSlice` in the `GzipSlice`, which does not seem necessary.

@olenagerasimova but it'll be a lot of boilerplate code, since gzip encoding can be used **only if** client sends `accept-encoding: gzip`, and there are no valid scenarios to use it without content-negotiation.

[...]

PR #405, opened 2021-11-22
Added `GzipSlice`

olenagerasimova:
Part of #404 
Added `GzipSlice` and test.


src/main/java/com/artipie/http/slice/GzipSlice.java
```diff
@@ -0,0 +1,141 @@
...
+     * @param connection Connection
+     * @return Completable action
+     */
+    private CompletionStage<Void> gzip(final Key key, final Connection connection) {
+        final CompletionStage<Void> future;
+        final CompletableFuture<Void> tmp;
+        try (PipedOutputStream resout = new PipedOutputStream();
+            PipedInputStream oinput = new PipedInputStream();
+            PipedOutputStream tmpout = new PipedOutputStream(oinput)
+        ) {
```
genryxy, line 110: is it possible that this streams will be closed before the target operation is finished? Because the result of this operation is `CompletionStage` and it seems that these streams could be closed earlier that it should be

genryxy (changes requested):
@olenagerasimova thanks, one question

src/main/java/com/artipie/http/slice/GzipSlice.java
genryxy: but `tmpout` and `resout` are used in one of the stage of `CompletableFuture`. Probably I am wrong and it does not matter in this case

olenagerasimova: These are piped input/output streams, and they are being processed when we gzip the content. After gzip workflow is fully done (input/output streams read/written fully), there is nothing to do in corresponding piped stream, that are used in the futures, so these streams can be closed.

genryxy: Okay, although I don't fully understand why after finishing gzip all workflow is fully done. In this case reading in `GZIPOutputStream` looks like a blocking operation therefore it should be wrapped in future

olenagerasimova: Can you just pass this PR to next review stage, please? The arc will decide is there smth wrong or not

genryxy (commented):
@g4s8 could you please take a look

olenagerasimova:
@g4s8 ping

g4s8 (changes requested):
@olenagerasimova gzip slice should encapsulate another sluce, not storage. It compress the response body of some origin slice which may produce dynamic content. Also, it keeps origin headers and adds content-encoding header to it.

olenagerasimova:
@g4s8 done, please, take a look

src/main/java/com/artipie/http/slice/GzipSlice.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public Response response(final String line, final Iterable<Map.Entry<String, String>> headers,
+        final Publisher<ByteBuffer> body) {
+        return connection -> this.origin.response(line, headers, body).send(
+            (status, rsheaders, rsbody) -> GzipSlice.gzip(connection, status, rsbody, rsheaders)
+        );
+    }
+
+    /**
+     * Gzip storage item.
```
kirill-h5i, line 57: > storage item

The javadoc is wrong here

kirill-h5i (changes requested):
@olenagerasimova just one minor comment about javadoc

olenagerasimova:
@kirill-h5i thanks, corrected java doc

olenagerasimova:
@g4s8 please, take a look/merge

g4s8 (approved):
@olenagerasimova thanks

olenagerasimova:
@genryxy please, approve

genryxy (approved):
@olenagerasimova thanks, sorry for wrong suggestions

PR #415, opened 2021-12-25
feat: return a list of the files present in a directory

### LINKED ISSUE #158: Add a SliceListing that lists file in a directory
This is needed for artipie/maven-adapter#87.

We need a slice that, given a key in the storage, will return a list of the files present in that directory in the form of a simple HTML listing.

We will also maybe need a `SliceRoute` that allow to dispatch requests depending on the fact a key is a directory or a file, for example to be able to say that GET requests get either dispatcher to `SliceDownload` or `SliceListing`. See https://github.com/artipie/http/issues/158#issuecomment-633629631 below for the proposed design.

dgarus:
Closes #158

- added `SliceListing` to lists blobs contained in given path
- added tests for `ListingFormat`
- added `todo` to test `SliceListing`
- added `todo` to implement `HTML` format


dgarus:
@g4s8 could you please have a look?

src/main/java/com/artipie/http/slice/SliceListing.java
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * Mime type.
+     */
+    private final String mtype;
+
+    /**
+     * Collection of keys to string transformation.
+     */
+    private final Function<Collection<? extends Key>, String> format;
```
g4s8, line 66: Does it work for `Function<List<Key>, String>`? Maybe we should change it to `Function<? extends Collection<? extends Key>>, String>`?

src/main/java/com/artipie/http/slice/SliceListing.java
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * Slice by key from storage.
+     *
+     * @param storage Storage
+     * @param standard Standard format
+     */
+    public SliceListing(
+        final Storage storage,
+        final SliceListing.Standard standard
```
g4s8, line 76: I don't think it's a good idea to have such a string coupling for enum type here, since it's not really extensible.

src/main/java/com/artipie/http/slice/SliceListing.java
```diff
@@ -0,0 +1,215 @@
...
+                                keys -> {
+                                    final String text = this.format.apply(keys);
+                                    return new RsFull(
+                                        RsStatus.OK,
+                                        new Headers.From(
+                                            new ContentType(this.mtype)
+                                        ),
+                                        new Content.From(
+                                            text.getBytes(
+                                                StandardCharsets.UTF_8
```
g4s8, line 138: Encoding should be determined with HTTP [content-negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation):
 - parse encoding from `accept` header if present or use default
 - add encoding to `Content-type` response header
 - encode HTTP body with this encoding

src/main/java/com/artipie/http/slice/SliceListing.java
```diff
@@ -0,0 +1,215 @@
...
+                )
+        );
+    }
+
+    /**
+     * Standard format implementations.
+     *
+     * @since 1.1.0
+     */
+    enum Standard implements Function<Collection<? extends Key>, String> {
```
g4s8, line 154: It's not a correct place for this enum (`SliceListing.Standard` - means standard implementations of slice listing slice?). We either need a type interface for format and put enum as nested type here as `ListingFormat.Standard` or create separate type `StandardListingFormats`

src/main/java/com/artipie/http/slice/SliceListing.java
```diff
@@ -0,0 +1,215 @@
...
+            "application/json",
+                keys -> Json.createArrayBuilder(
+                    keys.stream().map(Key::string).collect(Collectors.toList())
+                ).build().toString()
+        );
+
+        /**
+         * Mime type.
+         */
+        private final String mtype;
```
g4s8, line 180: I don't agree that mime keeping a mime type is a responsibility of listing output format:
 1. We may want to return the same format with different types (e.g. `application/json` if `Accept` header is `application/json` or `application/octet-stream` if `Accept` header is `application/octet-stream`)
 2. It's related to to HTTP layer (slice) to negotiate the correct `Content-type` response header, so it may want to use some additional mime-type parameters (like `charset` or others: `Content-type: application/json; charset=utf-8`)

g4s8 (changes requested):
@dgarus thanks, please check my comments

src/main/java/com/artipie/http/slice/SliceListing.java
dgarus: @g4s8 
Do we really need to support the `Accept-Charset` header?
The [documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset) contains the warning: Do not use this header. Browsers omit this header and servers should ignore it.

src/main/java/com/artipie/http/slice/SliceListing.java
dgarus: I did a separate type ListingFormat with enum.

src/main/java/com/artipie/http/slice/SliceListing.java
g4s8: No, the charset can be included into the mime type which is present by `Content-Type` header, e.g. `Content-type: application/json;encoding=utf-8`

dgarus: Thanks, I got it.
Added todo to implement Accept-Encoding header.

dgarus:
@g4s8 hello!
Could you please have a look?

src/main/java/com/artipie/http/slice/SliceListing.java
g4s8: The origin comment here was only about mime-type encoding only, not about other accept values included in header: current problem that this code hard-codes UTF-8 charset encoding for response, but client may request another encoding for this particular media type, e.g. `Accept: application/json;encoding=ISO-885901` (some windows encoding) - and it will be an error if we respond with `Content-Type: application/json;encoding=UTF-8` and this may cause runtime errors on client side. Of course we can use `UTF-8` as a default encoding here (if client doesn't specify it in accept header).

g4s8 (changes requested):
Thanks, please check my comment. Also, there are a lot of warnings in this PR about test coverage, please transfer related test from files-adapter: https://github.com/artipie/files-adapter/blob/master/src/test/java/com/artipie/files/BlobListFormatTest.java

src/main/java/com/artipie/http/slice/SliceListing.java
dgarus: @g4s8 Oh, I understood the problem is misusing words 'charset' and 'encoding'.

UTF-8 is charset, so the expected charset could be pointed by the [Accept-Charset](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset) header. 
The header [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) does not contain a 'charset' part:
`Accept: <MIME_type>/<MIME_subtype>`

Response's `Content-Type` header may contain a 'charset' part.
We should support the `Accept-Charset` header to enable a custom charset. 

But the documentation contains the warning: 
`Do not use Accept-Charset header. Browsers omit this header, and servers should ignore it. `

That is why I asked Do we need to support the `Accept-Charset` header? 

If this is the case, I'll create a to-do: "Add support for the `Accept-Charset` header."

src/main/java/com/artipie/http/slice/GzipSliceListing.java
```diff
@@ -0,0 +1,45 @@
...
+import com.artipie.http.Slice;
+import java.util.function.Function;
+
+/**
+ * This slice lists blobs contained in given path
+ * and gzips the response content.
+ * @see SliceListing
+ * @since 1.1.1
+ */
+public class GzipSliceListing extends Slice.Wrap {
```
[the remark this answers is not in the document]
dgarus: I deleted GzipSliceListing.

src/main/java/com/artipie/http/slice/SliceListing.java
g4s8: @dgarus I see now, thanks, if `Accept` can't contain charset MIME-param and `accept-charset` is obsolete, then we don't really need to negotiate client's encoding. Sorry for misunderstanding.

g4s8 (approved):
@dgarus thanks

PR #416, opened 2021-12-30
feat: add new method RqParams.findValues that return all parameter va…

### LINKED ISSUE #413: Parse request line query parameter list
`RqParams` class returns `Optional<String>` for query parameters assuming that HTTP request line URI contains only one query param value for name. But  actually it may have many params with same name and different values. Let's add additional method to get all parameters for `RqParams`: `List<? extends String> values(String name)`

_Originally posted by @baudoliver7 in https://github.com/artipie/artipie/pull/979#r774647115_

Yuurga:
Add new method  List<? extends String> values(String name)  for RqParams class. Method return list of all parameters values from a query. (#413)
resolves #413

### DIFF
--- a/src/main/java/com/artipie/http/rq/RqParams.java
+++ b/src/main/java/com/artipie/http/rq/RqParams.java
@@ -5,10 +5,14 @@
 package com.artipie.http.rq;
 
 import com.google.common.base.Splitter;
+
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URLDecoder;
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


g4s8:
@Yuurga thanks! CI pipeline was failed, please check error messages

genryxy:
@Yuurga thanks, I'd suggest to run locally `mvn qulice:check -Pqulice` to resolve codestyle issues

genryxy:
@Yuurga there are test [failures](https://github.com/artipie/http/runs/4767340810?check_suite_focus=true#step:5:1966) now

genryxy (approved):
@Yuurga thanks, looks good to me
@g4s8 please check

Issue #418, opened 2022-01-12
Redesign multipart filter method


g4s8:
Current design of `RqMultipart.filter` method leads to misunderstanding of its usage and unexpected behavior. Let's try to redesign it by providing new method and deprecating existing.

See this comment: https://github.com/artipie/pypi-adapter/pull/161#issuecomment-1010690896

PR #419, opened 2022-01-12
feat(multipart): inspect method for RqMultipart

g4s8:
Add `inspect` method instead of `filter` to inspect parts
one by one and either accept it or ignore. It accepts
`inspector` function which should make a decision on each part,
if user forget to expliticly accept or ignore element,
publisher will fail with exception - it will help to avoid
incorrect usage here. All ignored elements are filtered out and
are not present in downstream publisher.

---

Implementation

All accepted elements will be submitted to downstream publisher.
All ignored elements will be drained and filtered out then.
Inspector parameter can
modify part with restriction that it should either read it in body
or chain publisher in replaced version.
Downstream contains all accepted-by-inspector elements one by one.
Filter method was deprecated and implementation was changed to use
`inspect` method.

Fix: #418


g4s8:
CC: @artipie/members

src/test/java/com/artipie/http/rq/multipart/RqMultipartTest.java
```diff
@@ -210,6 +214,52 @@ void readOnePartOfRequest() {
...
+            ).inspect(
+                (part, inspector) -> {
+                    if (new RqHeaders(part.headers(), "Test").get(0).equals("2")) {
+                        inspector.accept(part);
+                    } else {
+                        inspector.ignore(part);
+                    }
+                    final CompletableFuture<Void> res = new CompletableFuture<>();
+                    res.complete(null);
+                    return res;
```
olenagerasimova, line 249: seems like these lines can be simplified to `CompletableFuture.completedFuture(null)`

olenagerasimova (approved):
@g4s8 thanks, please, check the comments

src/test/java/com/artipie/http/rq/multipart/RqMultipartTest.java
g4s8: Yes, it was initially `CompletableFuture.completedFuture(null)` but Java-8 don't have this method.

src/main/java/com/artipie/http/rq/multipart/RqMultipart.java
```diff
@@ -118,11 +148,124 @@ private String boundary() {
...
     }
+
+    /**
+     * A function which inspects upstream parts.
+     *
+     * @implNote it MUST either
+     * accept or ignore each part using sink downstream. In case if
+     * some part is not accepted or ignored, the publisher WILL fail
+     * with exception.
+     * @implNote Sink parameter is not thread safe - do not try to
```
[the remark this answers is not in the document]
g4s8: Thanks, fixed

olenagerasimova (approved):
@g4s8 thanks, looks good, please, release after merge

Issue #420, opened 2022-01-14
Integration tests for multipart


g4s8:
Add integration test for multipart request parser.

Related: artipie/pypi-adapter#163

PR #422, opened 2022-01-14
test(multipart): integration test for multipart

g4s8:
Add integration test for multipart using Vertx server deployment
and apache-httpclient as client. The test sends request with multipart
body and slice handler parses this request and save expected part body,
then test compares expected part with data sent.

Fixed the problem discovered during testing: some clients don't respect
`RFC6838` specification of media types by expecting **case-sensitive**
mime-type param values for boundary param (it's not case sensitive by spec).
To fix this, the feature g4s8/mime#36 was added to mime-type parser and
a new version v2.3.0 was released.

Close: #420

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,7 +104,7 @@ SOFTWARE.
     <dependency>
       <groupId>wtf.g4s8</groupId>
       <artifactId>mime</artifactId>
-      <version>v2.2</version>
+      <version>v2.3.2+java8</version>
     </dependency>
     <!-- Test only dependencies -->
     <dependency>
@@ -194,6 +194,17 @@ SOFTWARE.
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
       <version>4.5.13</version>
+    </dependency>
+    <!-- integration tests -->
+    <dependency>
+      <groupId>org.apache.httpcomponents.client5</groupId>
+      <artifactId>httpclient5</artifactId>
+      <version>5.1.2</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents.client5</groupId>
+      <artifactId>httpclient5-fluent</artifactId>
+      <version>5.1.2</version>
     </dependency>
   </dependencies>
   <build>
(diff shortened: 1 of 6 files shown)


olenagerasimova:
@g4s8 yes, I've checked it and made it fail:
- I've removed `AtomicReference` and saved the part to `InMemoryStorage`
- put some rpm archive into resources and used it as `data` instead of simple short text
- compare saved part from the storage and archive read from resources, they are different, data from the storage is larger

I can commit the changes if it would be convenient.

g4s8:
> @g4s8 yes, I've checked it and made it fail:
> 
>     * I've removed `AtomicReference` and saved the part to `InMemoryStorage`
> 
>     * put some rpm archive into resources and used it as `data` instead of simple short text
> 
>     * compare saved part from the storage and archive read from resources, they are different, data from the storage is larger
> 
> 
> I can commit the changes if it would be convenient.

@olenagerasimova can you please upload request payload data here?

olenagerasimova:
@g4s8 github does not allow to attach rpm files, so here is [a link to the file](https://github.com/artipie/rpm-adapter/blob/master/src/test/resources-binary/dbus-1.6.12-17.el7.x86_64.rpm) in rpm-adapter

g4s8:
> @g4s8 github does not allow to attach rpm files, so here is [a link to the file](https://github.com/artipie/rpm-adapter/blob/master/src/test/resources-binary/dbus-1.6.12-17.el7.x86_64.rpm) in rpm-adapter

@olenagerasimova thanks, reproduced

g4s8:
@olenagerasimova reproduced with plain string: https://github.com/artipie/http/pull/422/commits/028c5699076e149b2b0b25722d93914045f92de3 - see commit message for details.

g4s8:
@olenagerasimova should be fixed now: https://github.com/artipie/http/pull/422/commits/5daf8caedf45e5ebdf53de4ceca9ff9eaa65138f

olenagerasimova (approved):
@g4s8 thanks, let's merge and release

g4s8:
@olenagerasimova please check `v1.1.1` release with these changes

PR #427, opened 2022-01-22
feat: RsHasHeader able to check only a few headers among multiply ones.

### LINKED ISSUE #396: RsHasHeader can't check only one header among multiple ones
Presently, when the response contains multiple headers, we must pass all of them at once to RsHasHeader's ctor to make it match.

By example, if the response contains headers `Content-Type: application/json` and `Content-Length: 123`, if you check only `Content-Type: application/json`, it will fail. You have to check both.

dgarus:
Closes #396 

- changed `RsHasHeader` to check only a few headers.
- added a test for this case.


dgarus:
@g4s8 Hello!
Could you have a look, please?

g4s8 (approved):
@dgarus thanks

PR #433, opened 2022-02-03
test(multipart): multipart - storage integratin

g4s8:
Add integration test to verify multipart with
`FileStorage` integration. It performs upload
of big-enough binary data, parse this data from
multipart and save it to `FileStorage`, then
verify file's content.

Ref: #420
Ref: artipie/pypi-adapter#163


src/test/java/com/artipie/http/rq/multipart/MultipartITCase.java
```diff
@@ -183,6 +191,62 @@ void parseBigMultiparRequest() throws Exception {
...
+                            final CompletableFuture<Void> res = new CompletableFuture<>();
+                            res.complete(null);
+                            return res;
+                        }
+                    )
+                ).flatMap(part -> part).to(
+                    pub -> Single.just(new Content.From(pub))
+                ).to(SingleInterop.get()).thenCompose(
+                    content -> new FileStorage(path).save(new Key.From("data"), content)
+                ).thenApply(none -> StandardRs.OK)
```
olenagerasimova, line 218: This code does not reflect what should be done in python exactly: here the part is `flatMaped` into publisher, and then saved to storage, in python we also need to obtain header value from it. And that is why `flatMap(part -> part)` cannot be used there.

I'd suggest to extend binary part header with some info (other than `name`, add file name for example) and try to return it in the response.

olenagerasimova (changes requested):
@g4s8 please, check my comment

g4s8:
@olenagerasimova thanks, fixed your comment

g4s8:
> @g4s8 thanks, looks good, though I do not see how it helps with problem in python...

@olenagerasimova I tried this code many times, so now I'm sure that the problem is not in the integration of multipart and file-storage. So I'll check another components integration in pypi repo.

PR #435, opened 2022-02-16
fix: use shared thread pool for multipart

kirill-h5i:
Try to fix `RejectedExecutionException` in pypi adapter by using shared cached thread pool
instead of single thread executor per multipart processor, do not shutdown it on complete.

Ref: artipie/artipie#1000


g4s8:
@olenagerasimova this could fix artipie/artipie#1000 issue, you can try `v1.1.2-d` version (maybe we need local snapshots in Artipie for this kind of releases)

Issue #426, opened 2022-01-21
SliceListing.java:34-37: Test SliceListing when listing...


0pdd:
The puzzle `158-b912e34b` from #158 has to be resolved: 

https://github.com/artipie/http/blob/6a81452370b5318c5590d36639803595bfb4dc76/src/main/java/com/artipie/http/slice/SliceListing.java#L34-L37

The puzzle was created by Denis Garus on 21-Jan-22. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #451, opened 2022-06-06
test: add test for SliceListing

Swizbiz:
Add unit tests for text and json mime types

Close: #426

### DIFF
--- a/src/test/java/com/artipie/http/slice/SliceListingTest.java
+++ b/src/test/java/com/artipie/http/slice/SliceListingTest.java
@@ -18,28 +18,36 @@ import java.util.Collections;
 import javax.json.Json;
 import org.cactoos.map.MapEntry;
 import org.hamcrest.MatcherAssert;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
 
 /**
  * Test case for {@link SliceListingTest}.
- * @since 1.1.4
+ * @since 1.2
  */
 @SuppressWarnings("PMD.AvoidDuplicateLiterals")
 class SliceListingTest {
+    /**
+     * Storage.
+     */
+    private final Storage storage = new InMemoryStorage();
+
+    @BeforeEach
+    void setUp() {
+        this.storage.save(new Key.From("one/target1.txt"), new Content.Empty()).join();
+        this.storage.save(new Key.From("one/two/target2.txt"), new Content.Empty()).join();
+    }
 
     @ParameterizedTest
     @CsvSource({
-        "not-exists/, ''",
-        "one/, 'one/target1.txt\none/two/target2.txt'"
+        "not-exists/,''",
+        "one/,'one/target1.txt\none/two/target2.txt'"
     })
     void responseTextType(final String path, final String body) {
-        final Storage storage = new InMemoryStorage();
-        storage.save(new Key.From("one/target1.txt"), new Content.Empty()).join();
-        storage.save(new Key.From("one/two/target2.txt"), new Content.Empty()).join();
         MatcherAssert.assertThat(
-            new SliceListing(storage, "text/plain", ListingFormat.Standard.TEXT).response(
+            new SliceListing(this.storage, "text/plain", ListingFormat.Standard.TEXT).response(
                 get(path), Collections.emptyList(), Flowable.empty()
             ),
             new ResponseMatcher(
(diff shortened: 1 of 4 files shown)


Swizbiz:
@olenagerasimova please check PR

src/test/java/com/artipie/http/slice/SliceListingTest.java
```diff
@@ -0,0 +1,87 @@
...
+ * Test case for {@link SliceListingTest}.
+ * @since 1.1.4
+ */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+class SliceListingTest {
+
+    @ParameterizedTest
+    @CsvSource({
+        "not-exists/, ''",
+        "one/, 'one/target1.txt\none/two/target2.txt'"
```
olenagerasimova, line 35: How junit treats the space in such `CsvSource` format? Skips it as part of the separator or treats it as part of the value? As far as know, default separator is comma `,` and `''` (double single quotes) produces empty string and are required only to obtain this empty string value. So, to avoid any misunderstanding and confusion, I'd suggest to remove the spaces in separator and use single quotes only when required

src/test/java/com/artipie/http/slice/SliceListingTest.java
```diff
@@ -0,0 +1,87 @@
...
+import javax.json.Json;
+import org.cactoos.map.MapEntry;
+import org.hamcrest.MatcherAssert;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+/**
+ * Test case for {@link SliceListingTest}.
+ * @since 1.1.4
```
olenagerasimova, line 27: Usually we place only two digits into since value tag adding one to the second digit, so if current version is 1.1.3, then since will be 1.2. There is no need to fix it, it's just so you know for the future

src/test/java/com/artipie/http/slice/SliceListingTest.java
```diff
@@ -0,0 +1,87 @@
...
+                        String.format("application/json; charset=%s", StandardCharsets.UTF_8)
+                    ),
+                    new MapEntry<>("Content-Length", String.valueOf(json.length()))
+                ),
+                json.getBytes(StandardCharsets.UTF_8)
+            )
+        );
+    }
+
+    private static String get(final String path) {
```
olenagerasimova, line 84: I'd suggest to choose more verbose name for this method, for example `requestLine` or `rqLineFrom`, Yegor has a [blog post](https://www.yegor256.com/2014/04/27/typical-mistakes-in-java-code.html#method-names) about it, we try to follow his suggestions

src/test/java/com/artipie/http/slice/SliceListingTest.java
```diff
@@ -0,0 +1,87 @@
...
+                body.getBytes(StandardCharsets.UTF_8)
+            )
+        );
+    }
+
+    @Test
+    void responseJsonType() {
+        final String one = "one/example1.txt";
+        final String two = "one/two/example2.txt";
+        final Storage storage = new InMemoryStorage();
```
olenagerasimova, line 63: I'd suggest to extract `Storage` into private field and initialize it in `@BeforeEach` method, thus we avoid repeating initialization in each test

olenagerasimova (changes requested):
@Swizbiz thanks, added several minor comments, please, take a look

src/test/java/com/artipie/http/slice/SliceListingTest.java
Swizbiz: Junit skip space, but I remove it for common style.

Swizbiz:
> @Swizbiz thanks, added several minor comments, please, take a look

@olenagerasimova done

src/test/java/com/artipie/http/slice/SliceListingTest.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Test case for {@link SliceListingTest}.
+ * @since 1.2
+ */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+class SliceListingTest {
+    /**
+     * Storage.
+     */
+    private final Storage storage = new InMemoryStorage();
```
olenagerasimova, line 35: It would be better to do the initialization in `BeforeEach` method, thus each test method will work with new clean empty storage and we will never have some sort of data collision

olenagerasimova (changes requested):
@Swizbiz thanks, one more suggestion, please, take a look

Swizbiz:
> @Swizbiz thanks, one more suggestion, please, take a look

@olenagerasimova done

olenagerasimova (approved):
@Swizbiz thanks

PR #486, opened 2022-10-11
fix: add realm to WWW-Authenticate header

Swizbiz:
According to RFC7617, the 'realm' parameter is REQUIRED(https://www.rfc-editor.org/rfc/rfc7617.html#section-2)


dgarus (approved):
Thanks!

Issue #501, opened 2023-01-13
Artipie permissions model


olenagerasimova:
As we discussed, let's work on the permissions model. For now, we agreed on the following points:
- permissions will extend `java.security.Permission` and use, if possible, various `java.security` features (like `AllPermission`)
- each adapter should have possibility to implement its own permissions version
- users' permissions will be parsed/read from somewhere by `Policy` interface (or group of interfaces) and it should be possible to add custom `Policy` implementation by providing corresponding `jar` and some specific configuration 
-  users' permissions should be cached
- from the box we will first of all support yaml permissions and yaml policy. The format if the following: in the configuration storage each user will have his own yaml config with credentials, some info and roles:
```
security
├── roles
│   ├── java-dev.yaml
│   ├── admin.yml
│   ├── ...
├── users
│   ├── david.yaml
│   ├── jane.yaml
│   ├── ...
```
where `roles` contain an existing roles files (one file for one role) in the following format:
```yaml
# java-dev.yaml
java-dev: # role name
  artipie_basic_permission: # the name of the `java.security.Permission` implementation
    maven-repo: # repository name
      - read
      - write
    python-repo:
      - read
    npm-repo:
      - read
```
```yaml
# admin.yaml
admin:
  artipie_all_permission: {} # the name of the `java.security.Permission` implementation
```

User's yaml setting class has the following format:
```yaml
david:
  type: plain
  pass: qwerty
  email: jane@example.com # Optional
  enabled: true # optional default true
  roles:
    - java-dev
  permissions:
    artipie_basic_permission:
      rpm-repo:
        - read
```

PR #502, opened 2023-01-13
feat: permission interfaces skeleton

olenagerasimova:
part of #501 
Please, check issue description first.

Added some prototype of permissions interfaces set, config implementation. The upper level are some class like `com.artipie.asto.factory.Storages` to map permission name and factory and `Policy` interfaces set, which will load and parse permissions themselves. 

The permission itself is not yet implemented and I think that the exact implementation should be done after we agree on the whole infrastructure prototype.

The problem I see now is that adapter does not know concrete repository name (it's just not needed) and now we will have to pass it to demand some permission for some operation. Check the[ current simplest implementation](https://github.com/artipie/files-adapter/blob/98edeb588d0200561d034295eef872d2e2b0654b/src/main/java/com/artipie/files/FilesSlice.java#L80)

### DIFF
--- a/src/main/java/com/artipie/http/perms/ArtipieBasicPermission.java
+++ b/src/main/java/com/artipie/http/perms/ArtipieBasicPermission.java
@@ -1,0 +1,85 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2022 artipie.com
+ * https://github.com/artipie/http/blob/master/LICENSE.txt
+ */
+package com.artipie.http.perms;
+
+import java.security.Permission;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Set;
+import org.apache.commons.lang3.NotImplementedException;
+
+/**
+ * Artipie basic permission. This permission takes into account repository name and
+ * the set of actions. Both parameters are required, repository name is composite in the
+ * case of ORG layout {user_name}/{repo_name}.
+ * Supported actions are: read, write, delete. Wildcard * is also supported and means,
+ * that any actions is allowed.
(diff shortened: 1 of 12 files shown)


src/main/java/com/artipie/http/perms/ArtipieBasicPermission.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Required serial.
+     */
+    private static final long serialVersionUID = -2916496571451236071L;
+
+    /**
+     * Actions set.
+     */
+    private final Set<String> actions;
```
dgarus, line 34: Subclasses should always return actions in what they consider to be their canonical form (see javadoc for `java.security.Permission#getActions`). So, the best solution is using a mask approach and store actions as string.
Think we should use `FilePermission` as a base example to implement a common adapter permission.
An additional example of custom permission is on 5.2.1 The Permission Class (page 78) O'Reilly book

dgarus (changes requested):
@olenagerasimova thanks!
Please check my comments.

src/main/java/com/artipie/http/perms/ArtipieBasicPermission.java
olenagerasimova: Mask will be used when this class will be properly implemented, but I think it would be better to throw an exeption here

dgarus (approved):

dgarus:
> @dgarus thanks, renamed the classes and made actions string. Once again, I have no intention to check actions by string, in will be done with integer mask analogically with `FilePermission`.

Ok, thanks!

PR #504, opened 2023-01-17
feat: policy skeleton

olenagerasimova:
part of #501 

Added interfaces and some implementations for policies. `YamlPolicy` will be properly implemented in the next step.

Also, what do you say about renaming `StorageConfig` interface into `ArtipieConfig` or `BasicConfig` or `Config` and moving it into asto root? `PolicyConfig` is absolutely the same...

### DIFF
--- a/src/main/java/com/artipie/http/perms/Policies.java
+++ b/src/main/java/com/artipie/http/perms/Policies.java
@@ -9,7 +9,6 @@ import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import com.jcabi.log.Logger;
 import java.lang.reflect.InvocationTargetException;
-import java.security.Permission;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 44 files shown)


olenagerasimova:
@dgarus please, review

src/main/java/com/artipie/http/perms/Permissions.java
```diff
@@ -99,7 +99,7 @@ private static Map<String, PermissionFactory> init(final Map<String, String> env
                         if (existed != null) {
                             throw new ArtipieException(
                                 String.format(
-                                    "Storage factory with type '%s' already exists [class=%s].",
+                                    "Permission factory with type '%s' already exists [class=%s].",
```
dgarus, line 102:
```
private static final String DEFAULT_PACKAGE = "com.artipie.http";
```
This package looks strange for permissions.
Maybe create a package `com.artipie.security`?

src/main/java/com/artipie/http/perms/Policies.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Environment parameter to define packages to find policies factories.
+     * Package names should be separated with semicolon ';'.
+     */
+    public static final String SCAN_PACK = "POLICY_FACTORY_SCAN_PACKAGES";
+
+    /**
+     * Default package to find policies factories.
+     */
+    private static final String DEFAULT_PACKAGE = "com.artipie.http";
```
dgarus, line 34: This package looks strange for permissions.
Maybe create a package com.artipie.security?

src/main/java/com/artipie/http/perms/Policies.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+        return factory.getPolicy(config);
+    }
+
+    /**
+     * Finds and initiates annotated classes in default and env packages.
+     * @param env Environment parameters
+     * @return Map of {@link PolicyFactory}
+     */
+    private static Map<String, PolicyFactory> init(final Map<String, String> env) {
```
dgarus, line 77: This method is very similar to according method in the `Permissions` class.
We can transform this method to a separate class and use it to instantiate factories.

src/main/java/com/artipie/http/perms/Policy.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.security.PermissionCollection;
+
+/**
+ * Security policy.
+ *
+ * @param <C> Implementation of {@link PermissionCollection}
+ * @since 1.2
+ */
+public interface Policy<C extends PermissionCollection> {
```
dgarus, line 15: Policy should return `java.security.Permissions` – a heterogeneous collection of Permissions.

We have to give `com.artipie.http.perms.Permissions` a different name.

dgarus (changes requested):
@olenagerasimova thanks!
Please have a look at my comments.

src/main/java/com/artipie/http/perms/Policy.java
olenagerasimova: what is the problem with `java.security.PermissionCollection`? We can use any implementation we like, including `java.security.Permissions` which extends `PermissionCollection`. `java.security.Permissions` is final and not very convenient for us as we have roles. Most probably we will create another heterogeneous implementation of `PermissionCollection` which will deal with roles.

src/main/java/com/artipie/http/perms/Policies.java
olenagerasimova: I've tried to do it via interface with two parameters (one for factory object, one for annotation), but the problem is the class cast and `.value()` method call when we obtain type
```
final String type = Arrays.stream(clazz.getAnnotations())
                            .filter(ArtipiePermissionFactory.class::isInstance)
                            .map(inst -> ((ArtipiePermissionFactory) inst).value()) //this one
                            .findFirst()
                            .orElseThrow(
                                // @checkstyle LineLengthCheck (1 lines)
                                () -> new ArtipieException("Annotation 'ArtipiePermissionFactory' should have a not empty value")
                            );
```
annotation interface cannot extend other interfaces, thus there cannot be any common denominator between annotation interfaces.

olenagerasimova:
@dgarus thanks, created security package and moved classes. Also renamed `Permissions` and `Policies`

PR #505, opened 2023-01-20
feat: yaml policy and user permissions

olenagerasimova:
part of #501 

The skeleton for the new security policy and permissions is ready. New authorization slices with `z` in the name are added to replace the old ones without `z` later, suggested approach requires minimal changes in the adapters. `com.artipie.http.auth.Permission` and `com.artipie.http.auth.Permissions` interfaces and implementations are to be removed later.  

The missing part here is cashing for user's and roles' permissions, will be implemented later.


dgarus (changes requested):
@olenagerasimova thanks!
Please check my comments.

src/main/java/com/artipie/security/policy/YamlPolicy.java
```diff
@@ -68,7 +91,176 @@ public YamlPolicy(final BlockingStorage asto) {
     }
 
     @Override
-    public Permissions getPermissions(final String uname) {
-        throw new NotImplementedException("To implement");
+    public UserPermissions getPermissions(final String uname) {
+        final User urs = new User(this.asto, uname);
+        return new UserPermissions(urs.perms(), urs.roles(), new Roles(this.asto));
```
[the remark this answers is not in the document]
olenagerasimova: Let's discuss it at the office.

olenagerasimova:
@dgarus please, take a look. The PR was updated

src/main/java/com/artipie/security/perms/AdapterBasicPermission.java
```diff
@@ -48,36 +91,168 @@ public AdapterBasicPermission(final String repo, final String strings) {
...
+    /**
+     * Calculate mask from action.
+     * @param actions The set of actions
+     * @return Integer mask
+     */
+    private static int maskFromActions(final Set<String> actions) {
+        int res = Action.NONE.mask();
+        if (actions.isEmpty() || actions.size() == 1 && actions.contains("")) {
+            res = Action.NONE.mask();
+        } else if (actions.contains(Action.ALL.names().iterator().next())) {
```
dgarus, line 176: Overcomplicated

```
if (actions.contains("*"))
```

dgarus (changes requested):
@olenagerasimova thanks!
Please have a look at my comments.

src/main/java/com/artipie/security/test/FreePermissions.java
```diff
@@ -0,0 +1,40 @@
...
+import java.security.PermissionCollection;
+import java.util.Collections;
+import java.util.Enumeration;
+import org.apache.commons.lang3.NotImplementedException;
+
+/**
+ * Free permissions implies any permission.
+ * @since 1.2
+ */
+public final class FreePermissions extends PermissionCollection {
```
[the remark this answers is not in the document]
olenagerasimova: Please, check the implementation from jdk. Java's `AllPermissionCollection` does not imply any permission BEFORE `add()` method was called. In our use cases `add()` method will not be called at all.

src/main/java/com/artipie/security/policy/Policy.java
```diff
@@ -15,6 +16,11 @@
  */
 public interface Policy<C extends PermissionCollection> {
 
+    /**
+     * Free policy for any user returns {@link FreePermissions} which implies any permission.
+     */
+    Policy<PermissionCollection> FREE = uname -> new FreePermissions();
```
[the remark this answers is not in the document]
olenagerasimova: hm, actually, I've used it in [the main code](https://github.com/artipie/artipie/pull/1238/files#diff-75ba0295cba8b6bade3b7b3e076c2bb660e615d75ac94631157413483a3a6541) to support the situation when artipie is working without any security policy at all. Support for such case is required as we can work in such mode now.

So, maybe I should remove this `test` package?

src/main/java/com/artipie/security/test/PolicyByUsername.java
```diff
@@ -0,0 +1,41 @@
...
+import com.artipie.security.perms.EmptyPermissions;
+import com.artipie.security.policy.Policy;
+import java.security.PermissionCollection;
+
+/**
+ * Policy implementation for test: returns {@link FreePermissions} for
+ * given name and {@link EmptyPermissions} for any other user.
+ * @since 1.2
+ */
+public final class PolicyByUsername implements Policy<PermissionCollection> {
```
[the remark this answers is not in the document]
olenagerasimova: We use such approach (allow all actions for specific user) in the tests in each adapter, the example is [here](https://github.com/artipie/files-adapter/pull/146/files#diff-b7967f22526a85faa437c09722fad023482ced1942a813642690d5ca76dd5225). So, there are two options: it will be copied into each adapter or we can create it one time here.

src/main/java/com/artipie/http/auth/AuthzSlice.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    public Response response(
+        final String line,
+        final Iterable<Map.Entry<String, String>> headers,
+        final Publisher<ByteBuffer> body
+    ) {
+        final Response response;
+        if (this.control.allowed(Authentication.ANY_USER)) {
```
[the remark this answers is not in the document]
olenagerasimova: this condition will work in the case when artipie is working in "no policy" mode, in this case we do not require authentication. This mode is used in all ITs in artipie and can be used by our users to try it out.

src/main/java/com/artipie/http/auth/AuthzSlice.java
```diff
@@ -0,0 +1,87 @@
...
+            response = this.origin.response(line, headers, body);
+        } else {
+            response = new AsyncResponse(
+                this.auth.authenticate(headers, line).thenApply(
+                    result -> result.user().map(this.control::allowed).map(
```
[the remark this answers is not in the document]
olenagerasimova: here we check if user have required permission, check `AccessControl` interface. Also, any suggestions to improve classes and interfaces names are welcome.

olenagerasimova:
@dgarus thanks, corrected and answered

src/main/java/com/artipie/security/policy/YamlPolicy.java
```diff
@@ -32,11 +54,17 @@
...
+ * Or for admin:
+ * <pre>{@code
+ * admin:
+ *   enabled: true # optional default true
+ *   permissions:
+ *     adapter_all_permission: {}
+ * }</pre>
+ * Role can be disabled with the help of optional {@code enabled} field.
+ * <p>User yaml format example:
+ * <pre>{@code
```
Swizbiz, line 67: David has an email listed as jane@example.com

olenagerasimova:
@dgarus @Swizbiz all fixed, thanks

olenagerasimova:
@dgarus please, review. I've added caching with guava cache. The invalidation will be required in the case when policy is updated via REST API. We can update the policy via rest api if and only if our own yaml policy `CachedYamlPolicy` is used (in other cases we just have no idea where, what and how to update), which means we need to check if the `Policy<C extends PermissionCollection>` is actually `CachedYamlPolicy`. So, it means we can perform class cast safely and call `invalidate()` method.

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
```diff
@@ -0,0 +1,398 @@
...
+     * 1) supplier for user individual permissions
+     * 2) supplier for user roles
+     * 3) function to get permissions of the role.
+     * @param uname Username
+     * @return Callable to create {@link UserPermissions}
+     */
+    private Callable<UserPermissions> createUserPermissions(final String uname) {
+        return () -> new UserPermissions(
+            new UncheckedSupplier<>(
+                () -> this.users.get(uname, () -> new User(this.asto, uname).perms())
```
dgarus, line 191: Think, we could create `new User(this.asto, uname)` only ones to get user permission and a set of roles.

src/main/java/com/artipie/security/perms/UserPermissions.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @Override
+    public void add(final Permission permission) {
+        this.perms.get().add(permission);
+    }
+
+    @Override
+    public boolean implies(final Permission permission) {
+        boolean res;
+        final String ref = this.last.get();
```
dgarus, line 81: This method is not thread safe because of `this.last` state field

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
```diff
@@ -0,0 +1,398 @@
...
+            res = EmptyPermissions.INSTANCE;
+        } else {
+            res = new Permissions();
+            for (final String type : all.keys().stream().map(item -> item.asScalar().value())
+                .collect(Collectors.toSet())) {
+                final YamlMapping perms = all.yamlMapping(type);
+                if (perms == null || perms.keys().isEmpty()) {
+                    res.add(
+                        FACTORIES.newObject(
+                            type, new PermissionConfig.Yaml(Yaml.createYamlMappingBuilder().build())
```
dgarus, line 244: `new PermissionConfig.Yaml(Yaml.createYamlMappingBuilder().build())` can be a constant.

dgarus (changes requested):
Thanks!
Please take a look at my comments.

src/main/java/com/artipie/security/perms/AdapterBasicPermission.java
```diff
@@ -48,36 +91,168 @@ public AdapterBasicPermission(final String repo, final String strings) {
...
+         */
+        AdapterBasicPermissionCollection() {
+            this.perms = new ConcurrentHashMap<>(5);
+            this.any = false;
+        }
+
+        @Override
+        public void add(final Permission permission) {
+            if (permission instanceof AdapterBasicPermission) {
+                this.perms.put(permission.getName(), permission);
```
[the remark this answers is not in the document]
olenagerasimova: `isReadOnly()` check added, `getCanonicalName()` method exists in java's `BasicPermission` only, is final and has nothing to do with our `AdapterBasicPermission`. As for class name check - `AdapterBasicPermission` is final, no point to check.

olenagerasimova:
@dgarus as we discussed, I've reduced the amount of caches and joined user individual permissions and roles.

olenagerasimova:
@dgarus ping, please review

src/test/java/com/artipie/security/perms/AdapterBasicPermissionTest.java
```diff
@@ -0,0 +1,180 @@
...
+import org.hamcrest.core.IsEqual;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/**
+ * Test for {@link AdapterBasicPermission}.
+ * @since 1.2
+ */
+@SuppressWarnings({"PMD.AvoidDuplicateLiterals", "PMD.TooManyMethods"})
+class AdapterBasicPermissionTest {
```
Swizbiz, line 17: Since we can use not only the "read" action, but also "r", "download", "install", "pull". I think we should add some tests with them. 
And the same with the WRITE("write", "w", "publish", "push", "deploy", "upload") and DELETE("delete", "d", "remove") actions.

src/test/java/com/artipie/security/policy/CachedYamlPolicyTest.java
```diff
@@ -0,0 +1,224 @@
...
+    @BeforeEach
+    void init() {
+        this.asto = new BlockingStorage(new InMemoryStorage());
+        this.cache = CacheBuilder.newBuilder().build();
+        this.user = CacheBuilder.newBuilder().build();
+        this.roles = CacheBuilder.newBuilder().build();
+    }
+
+    @Test
+    void aliceCanReadFromMavenWithJavaDevRole() {
```
Swizbiz, line 59: I think we can give Alice 2 roles and check the size of the relevant caches as 2.

Swizbiz (commented):
@olenagerasimova Thank you!
Please, check my comments.

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
```diff
@@ -0,0 +1,391 @@
...
+         * @return Permissions of the role
+         */
+        public PermissionCollection perms(final String role) {
+            PermissionCollection res;
+            final String filename = String.format("roles/%s", role);
+            try {
+                final YamlMapping mapping = CachedYamlPolicy.readFile(this.asto, filename)
+                    .yamlMapping(role);
+                final String enabled = mapping.string(AstoRoles.ENABLED);
+                if (Boolean.FALSE.toString().equalsIgnoreCase(enabled)) {
```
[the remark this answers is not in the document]
olenagerasimova: Unfortunately no as `enabled` is nullable and `null` means that optional field is missing and user is enabled. 
(When parsing booleans null means false)

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
```diff
@@ -0,0 +1,391 @@
...
+            }
+            return roles;
+        }
+
+        /**
+         * Is user enabled?
+         * @return True is user is active
+         */
+        boolean disabled() {
+            return Boolean.FALSE.toString().equalsIgnoreCase(this.yaml.string(AstoRoles.ENABLED));
```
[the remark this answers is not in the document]
olenagerasimova: same here

src/test/java/com/artipie/security/perms/AdapterBasicPermissionTest.java
olenagerasimova: I've added separate test for `Action.Standard`, this one looks like complicated enough already

src/test/java/com/artipie/security/policy/CachedYamlPolicyTest.java
olenagerasimova: Added another user with two roles and negative assertion. The order of checking roles is not guarantied, so in the positive case we can never load all the roles.

olenagerasimova:
@Swizbiz thanks, corrected and answered

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
```diff
@@ -0,0 +1,391 @@
...
+         */
+        private final YamlMapping yaml;
+
+        /**
+         * Ctor.
+         * @param asto Storage to read user yaml file from
+         * @param username The name of the user
+         */
+        AstoUser(final BlockingStorage asto, final String username) {
+            this.yaml = getYamlMapping(asto, username);
```
dgarus, line 331: I think we should read `perms` and `roles` in the `ctor`. 
They have to be read only once. There is no sense to recreate `PermissionCollection` or  `Set<String>` for every method call.

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
```diff
@@ -0,0 +1,391 @@
...
+         * Storage to read roles yaml file from.
+         */
+        private final BlockingStorage asto;
+
+        /**
+         * Ctor.
+         * @param asto Storage to read roles yaml file from
+         */
+        AstoRoles(final BlockingStorage asto) {
+            this.asto = asto;
```
dgarus, line 280: I think we should read `perms` in the `ctor`.
They have to be read only once. There is no sense to recreate `PermissionCollection` for every method call.

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
```diff
@@ -0,0 +1,391 @@
...
+     * @param uname Username
+     * @return Callable to create {@link UserPermissions}
+     * @checkstyle LocalFinalVariableNameCheck (10 lines)
+     */
+    private Callable<UserPermissions> createUserPermissions(final String uname) {
+        final User astoUser = new AstoUser(this.asto, uname);
+        return () -> new UserPermissions(
+            new UncheckedSupplier<>(
+                () -> this.users.get(
+                    uname, () -> new User.Simple(astoUser.roles(), astoUser.perms())
```
dgarus, line 190: We have to add to the cache `new AstoUser(this.asto, uname)` like we do with `AstoRoles`

dgarus (changes requested):
Thanks!
Please take a look at my comments.

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
olenagerasimova: they are read only once, when requested role is not present in the cache.

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
olenagerasimova: I've correct the code, user yaml will be read and parsed only once in the case when `this.users` cache does not have corresponding user

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
olenagerasimova: agree, the instance of this class should be created only in `this.users` cache callable parameter

olenagerasimova:
@dgarus answered and corrected

olenagerasimova:
@dgarus done as you asked and from my point of view it's just disgusting and all the point of "custom user policy implementation" is nearly lost as now `UserPermissions` depends on some `AstoUser` which requires specific yamls

dgarus:
> 

Yes, you are right, interface `User` as parameter for `UserPermissions ` is useful.

Can we return it?

Sorry

olenagerasimova:
@dgarus it seems ok and convenient to use already existing `Function` interface to provide role's `PermissionCollection` by its name

dgarus:
> @dgarus it seems ok and convenient to use already existing `Function` interface to provide role's `PermissionCollection` by its name

In that case, we really could replace `AstoRole` by a static method.

olenagerasimova:
@dgarus done

Swizbiz (approved):
Thank you.

PR #506, opened 2023-01-20
feat: used `FactoryLoader` from asto

olenagerasimova:
part of #501 

used `FactoryLoader` from asto to avoid code duplication


olenagerasimova:
@dgarus could you please review this PR? This one should be clear, all the job regarding factories loading is done.

dgarus (approved):
Thanks!
lgtm

PR #513, opened 2023-02-17
feat: implement `Cleanable` in `CachedYamlPolicy`

olenagerasimova:
part of #501 
Corrected `CachedYamlPolicy` to implement `Cleanable` interface from `asto` to invalidate caches values.
Also, as we discussed, extracted `AuthUser` class from `Authentication` interface, removed groups as we do not need them anymore and used the class as method parameter in `Policy`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,7 +94,7 @@ SOFTWARE.
     <dependency>
       <groupId>com.artipie</groupId>
       <artifactId>asto-core</artifactId>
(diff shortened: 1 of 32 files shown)


olenagerasimova:
@dgarus please, review. As we discussed today, I've made class `AuthUser` final and added context field

dgarus (approved):
Thanks! lgtm

PR #517, opened 2023-03-01
feat: rename policy

olenagerasimova:
part of #501 

Renamed policy


olenagerasimova:
@dgarus please, check. I've just renamed policy from `artipie_policy` into `artipie` as we discussed before the weekend

dgarus (approved):
thanks!

PR #518, opened 2023-03-02
fix: perm for any repo

olenagerasimova:
part of #501 

Added fix and tests for the case with `adapter_basic_permission` for any repository.


olenagerasimova:
@dgarus please, check, that's not big but important fix

src/test/java/com/artipie/security/policy/CachedYamlPolicyTest.java
```diff
@@ -347,4 +406,48 @@ private byte[] tester() {
             "        - download"
         ).getBytes(StandardCharsets.UTF_8);
     }
+
```
dgarus, line 409: all these methods can be replaced by one
```
private byte[] userConfig(String user, String repo){
        String template = String.join(
            "\n",
            "%s:",
            "  permissions:",
            "    adapter_basic_permission:",
            "      %s:",
            "        - download"
        );
        return String.format(template, user, repo).getBytes(StandardCharsets.UTF_8);
    }
    
```

dgarus (changes requested):
@olenagerasimova thanks!
Please check my comment.

Do we really need a username in the permissions file? 
It looks like an info duplication because we already have username as file name.

olenagerasimova:
@dgarus thanks, corrected

dgarus (approved):
thanks!

PR #523, opened 2023-03-22
feat: factory should return collection

olenagerasimova:
part of #501 

Docker permissions format includes resource (images name) and thus has more levels in yaml setting format:
```yaml
docker_repository_permission:
  my-docker:
    my-alpine: # resource (image) name
      - pull
    ubuntu-slim:
      - pull
      - push
  docker-global:
    *:
      - pull
docker_registry_permission:
  my-docker:
    - catalog
    - base
```

To parse such config and create permission we need to parse this yaml differently as yaml for `adapter_basic_permission`. In order to do so, I've made two changes:
- `PermissionFactory` now can return permissions collection
-  `PermissionConfig` permission config can return sub-config

In other words, all the changes are required as there can be only one `AdapterBasicPermission` for one repository, but in docker there can be several `DockerPermission` for one repository.
[Why there is permissions type can be "repository" or "registry"](https://docs.docker.com/registry/spec/auth/scope/).

### DIFF
--- a/src/main/java/com/artipie/security/perms/PermissionConfig.java
+++ b/src/main/java/com/artipie/security/perms/PermissionConfig.java
@@ -11,9 +11,10 @@ import java.util.stream.Collectors;
 
 /**
  * Permission configuration.
+ * @param <T> Configuration original format type: yaml, json, database table, string etc.
  * @since 1.2
  */
-public interface PermissionConfig {
+public interface PermissionConfig<T> {
 
     /**
      * Permission name. The name can vary from the permission model,
@@ -39,13 +40,20 @@ public interface PermissionConfig {
     Set<String> sequence(String key);
 
     /**
+     * Get config in its original form: it can be yaml, json, string or
+     * any other representation from what config is parsed.
+     * @return Original form
+     */
+    T original();
+
+    /**
      * Yaml permission config.
      * Implementation note:
      * Yaml permission config allows {@link AdapterBasicPermission#WILDCARD} yaml sequence.In
      * yamls `*` sign can be quoted. Thus, we need to handle various quotes properly.
      * @since 1.2
      */
-    final class Yaml implements PermissionConfig {
+    final class Yaml implements PermissionConfig<YamlMapping> {
 
         /**
          * Yaml mapping to read permission from.
@@ -89,6 +97,11 @@ public interface PermissionConfig {
                 ).collect(Collectors.toSet());
             }
             return res;
+        }
+
+        @Override
+        public YamlMapping original() {
+            return this.yaml;
         }
 
         /**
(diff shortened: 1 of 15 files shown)


olenagerasimova:
@dgarus thanks, corrected as we discussed. But it was more convenient to make `PermissionConfig` extend `Config` from `asto` and add extra method to obtain keys (which are repository names in our case) collection.

olenagerasimova:
@dgarus please, review. You can also check https://github.com/artipie/docker-adapter/pull/551 to see how permissions are implemented in docker.

src/main/java/com/artipie/security/perms/PermissionConfig.java
```diff
@@ -91,6 +65,38 @@ public Set<String> sequence(final String key) {
...
+            }
+            return res;
+        }
+
+        /**
+         * Find wildcard key as it can be escaped in various ways.
+         * @param key The key
+         * @return Escaped key to get sequence or mapping with it
+         */
+        private Scalar getWildcardKey(final String key) {
```
andpopov, line 90: I see the value of parameter 'key' is always '*'
Can we get rid of it?

src/main/java/com/artipie/security/perms/PermissionConfig.java
```diff
@@ -91,6 +65,38 @@ public Set<String> sequence(final String key) {
...
+        @Override
+        public Set<String> keys() {
+            return this.yaml.keys().stream().map(node -> node.asScalar().value())
+                .map(Yaml::cleanName).collect(Collectors.toSet());
+        }
+
+        @Override
+        public PermissionConfig config(final String key) {
+            final PermissionConfig res;
+            if (AdapterBasicPermission.WILDCARD.equals(key)) {
```
andpopov, line 77: Here code uses equation to compare key with "*", but on line 92 code uses String.contains("*")

src/main/java/com/artipie/security/perms/PermissionConfig.java
olenagerasimova: no, we can not. Wildcard in yaml can be escaped in various ways: "*", '*', \*. So when just * is passed as a key we need to find actual key with escape signs in yaml and get corresponding value. Check [this](https://github.com/artipie/http/blob/62c7fd04ba0ac5bdd96a145067d92d3121b0d50e/src/test/java/com/artipie/security/policy/CachedYamlPolicyTest.java#L287) test case for better understanding

src/main/java/com/artipie/security/perms/PermissionConfig.java
olenagerasimova: yes, because we always un-escape mapping key when providing it (check keys() method), thus here only clean '*' can be passed. 
This work-around with wildcards is made because eo-yaml does not un-escape * when it is mapping name.

andpopov (approved):
@olenagerasimova thank!

PR #524, opened 2023-03-29
Repository content filters support

andpopov:
Part of: https://github.com/artipie/artipie/issues/1254

Filters support for repository:
- Filter's yaml configuration format
- Filters API
- Globbing filter
- Regexp filter
- FilterSlice


src/main/java/com/artipie/http/filter/FilterSlice.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Ctor.
+     * @param origin Origin slice
+     * @param yaml Yaml mapping to read filters from
+     * @param loader FilterFactoryLoader
+     * @checkstyle HiddenFieldCheck (10 lines)
+     */
+    public FilterSlice(final Slice origin, final YamlMapping yaml,
+        final FilterFactoryLoader loader) {
```
olenagerasimova, line 41: hm, I'd suggest to make `FilterFactoryLoader` private static field. The factory is one and only, there is no need to pass it into ctor. Also such approach guarantee it will not be pointlessly created on each ctor call

src/main/java/com/artipie/http/filter/GlobFilter.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Ctor.
+     *
+     * @param yaml Yaml mapping to read filters from
+     */
+    public GlobFilter(final YamlMapping yaml) {
+        this.matcher = FileSystems.getDefault().getPathMatcher(
+            String.format("glob:%s", yaml.string("filter"))
+        );
```
olenagerasimova, line 33: hm... Is it really legal to use file system path matcher with URI?

src/main/java/com/artipie/http/filter/GlobFilter.java
```diff
@@ -0,0 +1,41 @@
...
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.util.Map;
+
+/**
+ * Glob repository filter.
+ *
+ * @since 1.2
+ */
+public final class GlobFilter implements Filter {
```
olenagerasimova, line 19: please, add some docs here with config format and examples. It will help with review and can be used in wiki later.

src/main/java/com/artipie/http/filter/RegexpFilter.java
```diff
@@ -0,0 +1,43 @@
...
+import com.artipie.http.rq.RequestLineFrom;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+/**
+ * RegExp repository filter.
+ *
+ * @since 1.2
+ */
+public final class RegexpFilter implements Filter {
```
olenagerasimova, line 17: same request about java doc

src/test/java/com/artipie/http/filter/FiltersTest.java
```diff
@@ -0,0 +1,89 @@
...
+    @Test
+    void allows() {
+        final Filters filters = FiltersTest.createFilters(
+            String.join(
+                System.lineSeparator(),
+                "include:",
+                "  - type: glob",
+                "    filter: **/com/acme/**",
+                "  - type: glob",
+                "    filter: **/com/artipie/**",
```
olenagerasimova, line 50: hm what if we change format this way:
```yaml
include:
  glob:
    - **/com/acme/**
    - **/com/artipie/**"
   regexp:
    - regexp1
    - regexp2
exclude:
  ...
```

It will shorter (as there will be no need to repeat "type: glob") and canonical with permission settiongs.

src/test/java/com/artipie/http/filter/GlobFilterTest.java
```diff
@@ -0,0 +1,108 @@
...
+ * Test for {@link GlobFilter}.
+ *
+ * @since 1.2
+ */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+class GlobFilterTest {
+    /**
+     * Request path.
+     */
+    private static final String PATH = "/mvnrepo/com/artipie/inner/0.1/inner-0.1.pom";
```
olenagerasimova, line 26: what about line with request parameters?

src/test/java/com/artipie/http/filter/RegexpFilterTest.java
```diff
@@ -0,0 +1,138 @@
...
+ * Test for {@link RegexpFilter}.
+ *
+ * @since 1.2
+ */
+@SuppressWarnings({"PMD.UseLocaleWithCaseConversions", "PMD.AvoidDuplicateLiterals"})
+class RegexpFilterTest {
+    /**
+     * Request path.
+     */
+    private static final String PATH = "/mvnrepo/com/artipie/inner/0.1/inner-0.1.pom";
```
olenagerasimova, line 26: same here, rq line can have parameters and other parts. I'd suggest to add corresponding tests...

olenagerasimova (changes requested):
@andpopov thanks, please, check my comments

src/test/java/com/artipie/http/filter/FiltersTest.java
dgarus: Maybe something like this?

```
include:
  - glob: **/com/acme/**
  - regexp: regexp1
  - glob: **/com/artipie/**"
  - regexp: regexp2
```

src/main/java/com/artipie/http/filter/FilterSlice.java
andpopov: add static instance to FilterSlice

src/main/java/com/artipie/http/filter/GlobFilter.java
andpopov: added documentation.
GlobFilter matches only path part of request.

src/test/java/com/artipie/http/filter/GlobFilterTest.java
andpopov: only path part of requestis used for matching

andpopov (commented):
@olenagerasimova Please review changes

src/test/java/com/artipie/http/filter/FiltersTest.java
andpopov: The FactoryLoader need some explicit property to detect type of filter, this is reason why 'type: glob' is explicit syntax.
For example RegExp filter support several properties:
Yaml format:
  type: regexp
  filter: regular_expression
  raw: true
  case_insensitive: true

  where
    'type' is mandatory with value 'regexp'.
    'filter' is mandatory and value contains regular expression for request matching.
    'full_uri' is optional with default with value 'false'
      and implies to match with full uri or path part of uri.
    'case_insensitive' is optional with default value 'false'
      and implies to ignore case in regular expression matching.

Filter can have own set of properties and it is unclear how to detect type without explicit named property.

olenagerasimova:
@andpopov thanks, note, please, that CI is failing. Generally, it's better to make sure all the checks passed before asking for  review. As for format, discuss it with Denis, please.

andpopov:
@olenagerasimova, @dgarus
Changed format and added priotiry property for filter in yaml format.
Please review

dgarus (approved):
Thanks!

olenagerasimova (approved):
@andpopov thanks

PR #525, opened 2023-03-29
fix: correct all permission factory

olenagerasimova:
part of #501 

Corrected `AdapterAllPermissionFactory`: it's necessary to add permission into java `AllPermissionCollection` in order the collection implies any permission. 

Also, as now `PermissionFactory` returns collection and in docker adapter permissions are called with S at the end, I've renamed 
`adapter_basic_permission` -> `adapter_basic_permissionS`


olenagerasimova:
@dgarus another small fix, please, check

dgarus (approved):
Thanks!

PR #528, opened 2023-04-06
fix: case with user *

olenagerasimova:
part of #501 

fixed case with user `*` in `CachedYamlPolicy` and removed old permissions-related classes.


dgarus (approved):
Thanks!

PR #529, opened 2023-04-10
feat: several permissions in `OperationControl`

olenagerasimova:
part of #501 

For docker-adapter it's necessary to allow `OperationControl` check several permissions. I've missed this when updating permissions earlier in docker, check [this slice](https://github.com/artipie/docker-adapter/blob/0ab12a6475230069d5e882f783eaf64c0586559c/src/main/java/com/artipie/docker/http/ManifestEntity.java#L277) (in the case, when manifest is absent, we allow operation with permission from **this** slice or with permission from **origin**)


olenagerasimova:
@dgarus please, check

PR #537, opened 2023-05-11
feat: process default roles

olenagerasimova:
part of https://github.com/artipie/artipie/issues/1289

Added default roles processing (check policy files structure in the origin issue, link provided above):
 
- default user role is obtained from authenticated user context (if present)
- this default role is also added to user roles list
- user individual permission file is not required, now it's possible to set some default permissions and users will be allowed to perform all the actions described in this default permissions 
-  generally, if some permissions file (user or group) is absent, no exception is thrown (only error is logged) and `EmptyPermissions` instance (which simply does not allow any actions) is used for this permissions


src/main/java/com/artipie/security/perms/UserPermissions.java
```diff
@@ -67,25 +67,41 @@ public final class UserPermissions extends PermissionCollection {
...
+     * (before user individual permissions),
+     * pass the group name with the reference to the ctor.
      */
     private final AtomicReference<String> last;
 
     /**
      * Ctor.
      * @param user User individual permissions and roles
      * @param rperms Role permissions
+     * @param reference Last reference
```
dgarus, line 80: Is this the default role name?

olenagerasimova: when it is passed to the ctor - yes. Then, it's the name of the role (or null if it was user individual permissions), which implied the permission, passed into the last `implies()` method call.

dgarus: This adds excess complexity.
I think, adding a default role to the user's role list is enough.

dgarus (changes requested):
Thanks!
Please look at my comment.

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
```diff
@@ -363,16 +364,20 @@ private static PermissionCollection perms(final YamlMapping yaml) {
         /**
          * Get user roles collection.
          * @param yaml Yaml to read roles from
+         * @param user Authenticated user
          * @return Roles collection
          */
-        private static Collection<String> roles(final YamlMapping yaml) {
-            Set<String> roles = Collections.emptySet();
+        private static Collection<String> roles(final YamlMapping yaml, final AuthUser user) {
+            Set<String> roles = new HashSet<>(1);
```
dgarus, line 371: We can create one `Set` less

```suggestion
            Set<String> roles = Collections.emptySet();
```

src/main/java/com/artipie/security/policy/CachedYamlPolicy.java
```diff
@@ -363,16 +364,20 @@ private static PermissionCollection perms(final YamlMapping yaml) {
...
             if (!AstoUser.disabled(yaml)) {
                 final YamlSequence sequence = yaml.yamlSequence("roles");
                 if (sequence != null) {
                     roles = sequence.values().stream().map(item -> item.asScalar().value())
                         .collect(Collectors.toSet());
                 }
+                if (user.authContext() != null && !user.authContext().isEmpty()) {
+                    roles.add(String.format("default/%s", user.authContext()));
```
dgarus, line 379:
```suggestion
                    final String role = String.format("default/%s", user.authContext());
                    if (roles.isEmpty()) {
                        roles = Collections.singleton(role);
                    } else {
                        roles.add(role);
                    }
```

dgarus (changes requested):
One minor comment

dgarus (approved):

PR #541, opened 2023-05-15
feat: add login header

### LINKED ISSUE #540: Add header with login
Related to https://github.com/artipie/artipie/issues/1291

On the adapter level we do not know the user who is uploading/downloading artifact. In order not to authenticate the user again from the header (basic or token auth) let's add login to the header right after authentication.

olenagerasimova:
closes #540 

On the adapter level we do not know the user who is uploading/downloading/removing artifact. In order not to authenticate the user again from the header (basic or token auth) I've add login to the header "artipie_login" right after authentication. Then, we will use this info in the adapters to add artifact metadata record.


dgarus (approved):
Thanks!

PR #543, opened 2023-05-25
feat: added db processing

olenagerasimova:

what was done:

1) added db package to initialize and insert artifacts records
2) `ArtifactRecord` is an interface which will be implemented in every adapter in some adapter-specific way. All the information can available in various places in different adapters and we will implement interfaces not to do any extra calculations
3) Added dummy queue which does nothing for the cases when artifacts db is not configured
4) added exception and error handling in `EventsProcessor` class


olenagerasimova:
@dgarus please, take a look. Let's discuss and then I'll add more tests

olenagerasimova:
@dgarus as we discussed removed all db connected classes, added `ArtifactEvent` dto.

olenagerasimova:
@dgarus ping

src/main/java/com/artipie/scheduling/ArtifactEvent.java
```diff
@@ -0,0 +1,239 @@
...
+
+        /**
+         * Artifact version.
+         */
+        private final String version;
+
+        /**
+         * Package size.
+         */
+        private final long size;
```
dgarus, line 172: IMHO, `Package size` and `uploaded time` shouldn't be included to `ArtifactId`.

olenagerasimova: hm, what if I rename `ArtifactId` unto `ArtifactInfo`?

dgarus: I'm not sure that we need `ArtifactId` or `ArtifactInfo`. Will we have an according constraint in the DB?

olenagerasimova: Why does constraint matter? Here I just grouped artifact-related data into one class, repository-related data into another class and owner name is left alone. Instances of these classes will be created in the adapters and main goal here is to make it convenient to create one instance in one place. Thus, `RepositoryId` instance will be more likely created in one place and `ArtifactInfo` in another, during or after repository update process.

dgarus: > Why does constraint matter?

It looks like we cannot have two records with the same `package id`, for example, `grpId` + `artifactId` + `version`.
To enforce this rule we could create a database constraint.


> grouped artifact-related data into one class, repository-related data into another class

I think these are extra classes which have no sense.

olenagerasimova:
@dgarus removed "extra" classes

dgarus (approved):
thanks!

PR #544, opened 2023-06-01
feat: different consumer instances for each job

olenagerasimova:

Changed `QuartsService` to accept several consumers, consumer instance per job. Also added size method for queue, will be used for monitoring


dgarus (approved):
LGTM

PR #548, opened 2023-06-23
feat: delete events and login header

olenagerasimova:
Split delete artifact events into two types: delete artifact by version and remove all versions of artifacts by name (required in npm-adapter). Added more various ctors and class to get value of the artipie login header.


dgarus (approved):
lgtm
Thanks!

PR #551, opened 2023-07-06
feat: added base job

olenagerasimova:

Added base job, which extends quartz job and have useful method to shut down the job. This class can be extended in the adapters to create jobs to process proxy(or new) repositories. 

`QuartzService` class was moved to artipie/artipie and removed from http.

Added class for proxy events with artifact key, repo name and owner login.

Extracted interface from `EventQueue`, such approach allows us to create what ever implementations needed: for tests and for the case, when writing data to db is off and queue should not actually collect some items.


src/main/java/com/artipie/scheduling/EventQueue.java
```diff
@@ -4,61 +4,130 @@
...
 /**
- * Events queue with {@link ConcurrentLinkedQueue} under the hood.
+ * Events queue.
  * Instance of this class can be passed where necessary (to the adapters for example)
  * to add data for processing into the queue.
  * @param <T> Queue item parameter type.
  * @since 1.3
  */
-public final class EventQueue<T> {
+public interface EventQueue<T> {
```
dgarus, line 20: Why do we need this interface? Can we use pure `Queue<T>`?

dgarus (changes requested):
Thanks!
Please look at my comment.

src/main/java/com/artipie/scheduling/EventQueue.java
olenagerasimova: we can, but with our own interface we can hide under it's hood anything we want, even not a queue, also we can extend it how we want and need. 
Also, if switch to queue now, we will need to correct a lot of adapter again.

olenagerasimova:
@dgarus thanks, removed queue interface. As for test queue implementation: we can simply use java `LinkedQueue`, no other implementations required

dgarus (approved):
Thanks!
lgtm

PR #553, opened 2023-08-01
feat: log upload/download events

olenagerasimova:
Part of https://github.com/artipie/artipie/issues/1291

Extended `SliceUpload` and `SliceDelete` to log artifact add/remove events into queue. This will be used in `files-adapter`, where these slices are responsible for add/remove data files.


dgarus (approved):
Thanks!
lgtm

PR #556, opened 2023-08-30
feat: added equals() to artifact events

olenagerasimova:

Added equals and hashCode methods to `ArtifactEvent` and `ProxyArtifactEvent`


dgarus (approved):
Thanks!

